PR #1, opened 2018-03-19
V3 rebase

martinstengel:


adamcpovey:
I've merged four of these commits into the trunk. We'd like to see some plots for the snow/ice albedo changes before merging those, so I'll keep this request open for now.

adamcpovey:
I feel like we should resolve this one. What do we think of the albedo numbers?

adamcpovey:
@garethethomas I think the changes to [correct_for_ice_snow.F90](../blob/master/pre_processing/correct_for_ice_snow.F90) are the only bits of this we might still want to keep. Do you have any opinions on the numbers used there for emissivity? I have none other than the desire to cite wherever they came from.

PR #18, opened 2019-04-30
Sunglint correction for SEVIRI

simonrp84:
Had enough of ORAC and its bugs, this is someone else's problem for once.

### DIFF
--- a/pre_processing/cox_munk.F90
+++ b/pre_processing/cox_munk.F90
@@ -1080,7 +1080,6 @@ subroutine cox_munk3_calc_shared_geo_wind(solza, satza, solaz, relaz, u10, v10, 
    real(kind=sreal) :: zeta, eta
    real(kind=sreal) :: cosomega
 
-
    !----------------------------------------------------------------------------
    ! Precalculate trigonometric functions
    !----------------------------------------------------------------------------
@@ -1098,9 +1097,9 @@ subroutine cox_munk3_calc_shared_geo_wind(solza, satza, solaz, relaz, u10, v10, 
 #else
    shared%sin_satza = sqrt(1. - shared%cos_satza * shared%cos_satza)
 #endif
-   shared%cos_relaz = cos((180. - relaz) * d2r)
+   shared%cos_relaz = cos(relaz * d2r)
 #ifdef COMPATIBILITY_MODE
-   shared%sin_relaz = sin((180. - relaz) * d2r)
+   shared%sin_relaz = sin(relaz * d2r)
 #else
    shared%sin_relaz = sqrt(1. - shared%cos_relaz * shared%cos_relaz)
 #endif
@@ -1243,7 +1242,6 @@ subroutine cox_munk3(i_band, shared_geo_wind, ocean_colour, rho)
    ! Sun-glint/Cox and Munk variables
    real(kind=sreal)            :: R_sf
    real(kind=sreal)            :: rhogl
-
 
    !----------------------------------------------------------------------------
    ! Calculate white-cap fraction (and ensure it is no greater than 1)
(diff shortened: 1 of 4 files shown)


adamcpovey:
That last image does look a lot more realistic. I can check the various regression tests but not in the next fortnight. Curious what Greg thinks.

gmcgarragh:
And so what about the lines in cox_munk.F90 that switch it the other way:

   ! Note that the relative azimuth used in Cox and Munk is the other way round
   ! from the convention used in the rest of ORAC. Here backscattering equates
   ! to a zero relative azimuth (i.e. if the satellite is looking away from the
   ! sun, azi = 0). Hence the 180 degree correction.
   rrelaz = d2r * (180. - relaz) ! relative azimuth

gmcgarragh:
What we know:  The main ORAC processor wants it like this:   RAA is 180 when sun and satellite are on opposite sides of the pixel, and RAA = 0 when sun and satellite are on the same side of the pixel.  This is standard in plane-parallel RT and is assumed throughout the main processor and in the lookup tables.  Let's call this the *DISORT* standard.  Let's call the opposite the *anti-DISORT* standard.

To solve this issue once and for all there needs to be a *strict* standard in the code.  With a strict standard one can eliminate the question at any point in the code of what standard is being used.   I would propose that the data readers put the relative azimuth into the DISORT standard *immediately* (some readers may need this conversion and some may not)  and that the RAA is propagated *throughtout* the rest of the ORAC system in the DISORT standard.  Then identify calculations that need it the other way and convert their RAA inputs right before the calculation.

The comment in the Cox-Munk code that I posted in my previous post implies that the input to Cox-Munk is in the anti-DISORT standard and therefore a conversion is made to the DISORT standard reqired by Cox-Munk.  This is ultimately incorrect.  So it seems that Simon's fix is putting the RAA in the anti-Disort standard so that the conversion in the Cox-Munk code puts it back the right way.  Both these conversions should be eliminated.

simonrp84:
I hadn't noticed the `rrelaz = d2r * (180. - relaz) ! relative azimuth` line. Agree that both it and my addition should be removed.
I also definitely agree that we should have a standard throughout the code, ensuring that everything is in DISORT standard seems, as you say, to be the best choice.

Today I've checked the three GEO sats, they were all in DISORT standard *but* I noticed that some had the azimuths reversed. So for a pixel _East_ of the satellite the satellite azimuth angle would be 90 degrees instead of 270, and the same with the solar azimuth. This had no effect for the RAA (as both were 'wrong') but I have corrected these for consistency (it will have an effect for RTTOV, for example).

Tomorrow I'll do some more tests and commit my changes (+ do as you suggest for Cox Munk) as part of this PR.

simonrp84:
Actually, @gmcgarragh that's not the problem.
The `rrelaz = d2r * (180. - relaz) ! relative azimuth` line is only in the `cox_munk` and `cox_munk2` functions, but the get_surface_reflectance.F90 code only calls `cox_munk3`.

I will update my PR to add the line to `cox_munk3` as well, and then remove my addition from get_surface_reflectance.F90

simonrp84:
This gets worse and worse! Lines 1101-1106 of cox_munk.F90:
```
   shared%cos_relaz = cos((180. - relaz) * d2r)
#ifdef COMPATIBILITY_MODE
   shared%sin_relaz = sin((180. - relaz) * d2r)
#else
   shared%sin_relaz = sqrt(1. - shared%cos_relaz * shared%cos_relaz)
#endif
```
So we're already, incorrectly, doing a flip of the RAA and my fix is just undoing that flip. Getting rid of the `180 - ` bit fixes the problem.

PR #26, opened 2020-04-21
Renaming NetCDF routines

adamcpovey:
This is mostly me testing how to create my own fork, commit to it, and then pull from that to the master.

The commits themselves correct the naming of routines in `orac_ncdf_m`, which are currently identical to some routines in the Fortran NetCDF library. This may be preventing compilation for some users.


pre_processing/netcdf_output_check.F90
```diff
@@ -39,7 +39,7 @@ subroutine netcdf_output_check(output_path, lwrtm_file, swrtm_file, prtm_file, &
    if (verbose) write(*,*) 'Albedo file: ', trim(alb_file)
    if (nf90_open(path=trim(adjustl(output_path))//'/'//trim(adjustl(alb_file)), &
```
simonrp84, line 40: Does this mean that we're now using the default `nf90_open` from the netcdf library itself rather than `nc_open` that came from our own `./common/` netCDF routines?
I'm curious as in some place you're using `nf90_open` and in other places you're using `ncdf_open`....

adamcpovey: I believe I only used `nf90_open()` in `netcdf_output_check()`; if it turns up anywhere else please flag that.

In that routine, we're  opening the file just to see if it exists. Using `nc_open()` there would cause the code to exit if the file was missing, which is exactly what this routine is attempting to prevent.

Now, I really don't like this routine. It's in a stupid `for` loop and a very blunt response to a specific user's difficult network infrastructure, but it seems DWD need it.

simonrp84: That's inelegant, but not your problem. Fair enough, I have no other comments on this. Thanks for the renaming!

PR #28, opened 2020-05-15
Add subsetting to SLSTR and fix it for SEVIRI

adamcpovey:
These haven't been through my usual rigorous quality control, but I feel I should get them up beore the weekend so Gareth can get on with his merge. They compile and run on my AOPP machine with all three compilers and introduce no regressions in my short tests (other than SEVIRI completely changing as the `relazi` field is now completely different).

Apparently my short test areas were somewhat poorly chosen. I'll deal with that later.

Also, I think I finally worked out how to run my own branch with `git cherry-pick`. We'll see how long it takes before I have to scrap my repo and make a new one...


pre_processing/read_seviri.F90
```diff
@@ -435,7 +443,7 @@ subroutine read_seviri_l1_5_nat_or_hrit(l1_5_file, imager_geolocation, &
    column0 = startx - 1
    column1 = startx - 1 + nx - 1
 
-   if (.not. hrit_proc .or. (nx .eq. 3712 .and. ny .eq. 3712)) then
+   if (.not. hrit_proc .and. (nx .eq. 3712 .and. ny .eq. 3712)) then
```
simonrp84, line 446: Why has this been changed?

pre_processing/read_seviri.F90
```diff
@@ -485,8 +510,11 @@ subroutine read_seviri_l1_5_nat_or_hrit(l1_5_file, imager_geolocation, &
       imager_angles%solzen(startx:,:,1)       = preproc%sza(column0+1:column1+1,line0+1:line1+1)
       imager_angles%solazi(startx:,:,1)       = preproc%saa(column0+1:column1+1,line0+1:line1+1)
       imager_angles%satzen(startx:,:,1)       = preproc%vza(column0+1:column1+1,line0+1:line1+1)
-      imager_angles%satazi(startx:,:,1)       = preproc%vaa(column0+1:column1+1,line0+1:line1+1)
+      !imager_angles%satazi(startx:,:,1)       = preproc%vaa(column0+1:column1+1,line0+1:line1+1)
       imager_measurements%data(startx:,:,:)   = preproc%data(column0+1:column1+1,line0+1:line1+1,:)
+
+      ! Offset subset for inversion of satazi
+      imager_angles%satazi(startx:,:,1)       = preproc%vaa(3712-column1:3712-column0,line0+1:line1+1)
```
simonrp84, line 517: This is hard-coded for full disk imagery (3712 columns), and will fail for super rapid scan mode.

pre_processing/read_slstr_funcs.F90
```diff
@@ -32,15 +32,21 @@
...
+! 2020/04/14, AP: Add subsetting to the read functions.
 !
 ! Bugs:
 ! SLSTR colocation is poor. Aerosol retrieval unusable. Cloud retrieval suspect.
 !-------------------------------------------------------------------------------
 
+! When defined, the offset between the nadir and oblique views is assumed
+! to be constant. Otherwise, the two longitude fields are read and compared
+! to determine an appropriate offset.
+#define CONSTANT_OBLIQUE_OFFSET 548
```
simonrp84, line 44: Where has this number come from? Could you add a reference?

pre_processing/read_slstr_main.F90
```diff
@@ -234,42 +221,51 @@ subroutine read_slstr(infile, imager_geolocation, imager_measurements, &
...
       else if (band_ids(i) .eq. 10) then
-        imager_measurements%data(:,:,i) = imager_measurements%data(:,:,i) * 0.92
+         imager_measurements%data(:,:,i) = imager_measurements%data(:,:,i) * 0.92
       else if (band_ids(i) .eq. 11) then
-        imager_measurements%data(:,:,i) = imager_measurements%data(:,:,i) * 0.93
+         imager_measurements%data(:,:,i) = imager_measurements%data(:,:,i) * 0.93
       else if (band_ids(i) .eq. 12) then
-        imager_measurements%data(:,:,i) = imager_measurements%data(:,:,i) * 0.93
+         imager_measurements%data(:,:,i) = imager_measurements%data(:,:,i) * 0.93
       end if
```
simonrp84, line 262: Can we have these defined somewhere obvious, like CONSTANT_OBLIQUE_OFFSET is, so that they're easy to change in the future? I worry that if, like before, they're buried here then people won't notice them.

pre_processing/read_seviri.F90
adamcpovey: The original code required all files that did not contain "H-000-MSG" to be processed using the memory efficient form. That is incompatible with the angle subsetting. By switching to `.and.` we only use the memory efficient mode for full-disc reads that don't contain that string.

pre_processing/read_seviri.F90
adamcpovey: True

pre_processing/read_slstr_funcs.F90
adamcpovey: It comes from my own analysis of SLSTR files as the modal offset found by the other code. I believe it was a Jupyter notebook, that I should be able to print as a PDF. That can't go in the code but it might go in the wiki.

pre_processing/read_slstr_main.F90
adamcpovey: They can be, but I thought the idea was the code should be deleted ASAP?

adamcpovey:
I agree that the subsetting issue would be better dealt with within the SEVIRI reader itself.

Having looked at the reader, I am *definitely* not doing that. Should we poke @gmcgarragh?

gmcgarragh:
Adam, your poke has been received.

You emailed me a while back about two problems:

1) Some subsetting indexing issue with the ORAC side of reading SEVIRI that I figured you just fixed. 

2) Something to do with geometry being different for subsetted data from the same block in the full disk.  From what I remember my regression tests for this show that subsetted data is identical to the same block in the full disk.  So I would like to know where this issue comes from.  If you are seeing this then it is an obvious bug.  Perhaps my regression tests are not covering all cases.

The azimuth angle thing with ORAC has been crazy.  I am pretty sure I had everything configured correctly after I stopped working on ORAC.  All I see are emails about it being wrong and never a reason why with solutions like flopping angles around.  Everything from the reader interfaces should be in and propagate through ORAC in the the PPRTM frame.  ORAC devs need to be sure not to get into situations where they loose track of the changes made to ORAC in this regard.

As far as changing seviri_util.  I designed it to be conform with the specs given by EUMSTAT.  If it is not compatible with the PPRTM geometry frame that ORAC wants then that should be corrected in the ORAC interface to the reader just like it is for other instruments.

If the subsetting is a problem then it needs to be fixed.  Please send me an example or Simon if you want to go at it then fix it, I think you have push privileges.

-Greg

simonrp84:
Hey @gmcgarragh , nice to 'see' you on here :-)
I agree the azimuth thing is a real headache, different bits of the code are using different definitions of azimuth.

Regarding SEVIRI, what is the PPRTM frame you mention? That's a new one on me and it's not in the EUM documentation. The SEVIRI azimuths are flipped relative to the common convention, but match what DISORT expects. However, this means that the surface code, RTTOV, etc get the angles wrong. We have agreed that ORAC should follow the expected convention and that we flip the angles where necessary (e.g: DISORT). Hence my suggestion that this should be implemented in the upstream `seviri_util` rather than in ORAC. If you feel that it's better to have `seviri_util` follow the DISORT definition of azimuths then we can fix it in ORAC instead, though, but it'd be good to put a comment in the `seviri_util` source code mentioning the azimuths don't follow the common convention but DISORT instead.

Cheers,
Simon

pre_processing/read_seviri.F90
simonrp84: I'll look into this more, I don't think it's doing what you expect it to. Or maybe it's not doing what I expect it to!

pre_processing/read_seviri.F90
simonrp84: Can you use the image sizes from `read_seviri_dimensions` instead of hardcoding `3712`? From memory that should contain the correct size.

pre_processing/read_slstr_funcs.F90
simonrp84: Ok, that's probably overkill. Can you just add to the comment saying you foudn this number empirically?

pre_processing/read_slstr_main.F90
simonrp84: Yeah, I agree that they should be...but IMO it's better to make this obvious, just in case the L1 calibration isn't fixed within the timespan of our memory ;-)

gmcgarragh:
Simon,

PPRTM is "Plane Parallel RTM".  It refers to the same geometry convention that DISORT, (V)LIDORT, XRTM, and other PPRTM's use.  It is also the convention that the ORAC RTM expects.  Sorry, I was trying to be general.  It is equivalent to saying "the DISORT convention".

I am confused by your statements:

"different bits of the code are using different definitions of azimuth."

and particularly the confusing:

"The SEVIRI azimuths are flipped relative to the common convention, but match what DISORT expects. However, this means that the surface code, RTTOV, etc get the angles wrong. We have agreed that ORAC should follow the expected convention and that we flip the angles where necessary (e.g: DISORT)."

I don't know what "common convention" and "expected convention" are.

I outlined my opinion on this a couple years ago.  Here is the basic idea:

Since ORAC RT expects the PPRTM convection then it should be propagated that way through ORAC.  This means converting convention in the ORAC interface to the readers, as close as possible to the reading, and making sure it is *that way* all the way through to the end.  If one of the end points is RTTOV then it should be converted to whatever the hell RTTOV wants at that point.

As far as seviri_util, it is not just an ORAC reader.  Amazingly, other people evidently use it.  It tries to adhere to EUMSTAT conventions.  If I am in error about these conventions then I would like to hear about that.  Whatever these conventions are, they should be converted by ORAC for its own purpose.  I am pretty sure I got this right from the beginning in the ORAC interface to seviri_util.

I am a bit discouraged as this problem has been rehashed for years and I can barely believe it is still happening.  It would be nice if we could build a test based on RT principles, using an artificially created idealised surface and atmospheric scenario, that can positively confirm things are right for each instrument.

-Greg

simonrp84:
> PPRTM is "Plane Parallel RTM". It refers to the same geometry convention that DISORT, (V)LIDORT, XRTM, and other PPRTM's use. It is also the convention that the ORAC RTM expects. Sorry, I was trying to be general. It is equivalent to saying "the DISORT convention".

Thanks.
 
> I don't know what "common convention" and "expected convention" are.
The 'common convention' is that used by the majority of satellite operators (EUM, NOAA, NASA) as well as most downstream products (MODIS BRDF) and RTTOV:
1) Pixel is due North of the satellite, VAA = 180 degrees.
2) Pixel is due East of the satellite, VAA = 270 degrees.
3) Pixel is due South of the satellite, VAA = 0 degrees.
4) Pixel is due West of the satellite, VAA = 90 degrees.

By 'expected convention' I meant the same thing, sorry for the lack of clarity. Don proposed that we use the above definition throughout ORAC.
> I outlined my opinion on this a couple years ago. Here is the basic idea:
> 
> Since ORAC RT expects the PPRTM convection then it should be propagated that way through ORAC. This means converting convention in the ORAC interface to the readers, as close as possible to the reading, and making sure it is _that way_ all the way through to the end. If one of the end points is RTTOV then it should be converted to whatever the hell RTTOV wants at that point.

The problem is, ORAC RT is the only bit of the code expecting it that way, and it's caused a great deal of confusion and errors in other parts of the code. Hence the recent effort by all of us to be consistent throughout. It seems to me that this is an issue where two fields come together: The RT people have one definition, the satellite ops and downstream people have another.

> As far as seviri_util, it is not just an ORAC reader. Amazingly, other people evidently use it. It tries to adhere to EUMSTAT conventions. If I am in error about these conventions then I would like to hear about that. Whatever these conventions are, they should be converted by ORAC for its own purpose. I am pretty sure I got this right from the beginning in the ORAC interface to seviri_util.

I've just dug through the EUM L1.0 processing code and their convention is as described above. Currently, seviri_util has the East-West direction flipped, so the VAA for an pixel due West of the spacecraft is given as 270 instead of 90.

> I am a bit discouraged as this problem has been rehashed for years and I can barely believe it is still happening. It would be nice if we could build a test based on RT principles, using an artificially created idealised surface and atmospheric scenario, that can positively confirm things are right for each instrument.

I fully agree. I've too much of my time in Oxford fixing bugs that, now that we've figured it out, have azimuth angles as their root cause! But the good news is that everyone (AOPP+RAL) now understands what the issue is and are working to fix it. We'll have to make sure that it's *very* well documented to avoid having a similar issue in the future.

pre_processing/read_seviri.F90
adamcpovey: I'm making that change but `read_seviri_dimensions` just returns 3712.

adamcpovey:
How's that?

gmcgarragh:
Simon,

Sorry for the delayed response.  Been finishing *unending* work for IWGGMS-16.

OK.  If that is the way it is now then so be it.  I still would have argued that the default ORAC convention should be the PPTRTM convention because that is what the RT at the ORAC core uses and decades of RT literature uses.  But, keeping to a standard, whatever it is is the most important.

I think I agree that seviri_util should be fixed now.  I did not realise that it did not conform to the documented convention.  I will break out my regression tests and look in to it in detail over the next week.

I hear you on the fixing bugs complaint.  Although endlessly fixing/improving ORAC may seem a bit tantalizing, try not to get to bogged down with software work.  There is a line in what we do that is between coding and actually doing science.

cheers,

-Greg

adamcpovey:
The intention is for all of ORAC to use the PPTRTM convention. For some reason, all of the satellite readers did the opposite. That's what we're trying to fix.

I try to keep to the coding necessary to get the results I want. Sadly, the bugs keep getting between me and usable data...

simonrp84:
Adam, that's not what we agreed. DISORT uses the opposite definition of relative azimuth angle, hence the need for flipping it 180.

Also, it's not "for some reason" that all the satellite readers have it this way, it's because that has been the convention for decades. As I said to Greg, it seems the RT people do things the opposite way around. It's unreasonable to expect satellite operators to define their angles for the (relatively small) RT community instead of all the other users.

simonrp84:
Thanks for your comments on this, @gmcgarragh. It's a pain that these two different conventions exist, but such is the way of things. Happy to help test seviri_util changes along the way, although unfortunately I won't have much time at the moment for actually helping with the coding.

Glad we're able to close this off. I've already corrected the angles in my Himawari code and the ORAC output for effective radius looks much more consistent across the day now. Really satisfying to see the effect!

adamcpovey:
Apologies - I was taking the PPRTM convention to be whatever was defined earlier in this thread as I can never keep straight who uses what. That was presumably you defining our convention.

By "satellite readers", I meant the code within ORAC that reads the angles. The satellite datasets can define whatever ruddy convention they like provided we convert it into what we want.

pre_processing/read_seviri.F90
simonrp84: This still is problematic, as it won't work in memory efficient mode for RSS images that have 1392 rows for VIS/IR channels, nor will it work for super RSS data that has differing row and columns counts.
Can this be refactored to take the above into account? We know the rows + columns requested by the user and we also (theoretically, even though it's broken just now) know the actual rows + columns in the data, so it should be a case of comparing the two.

simonrp84:
@adamcpovey The SEVIRI rows / columns issue is still troubling here, as it'll cause problems for non-fulldisk processing. Aside from that, the changes you made look OK.

pre_processing/read_seviri.F90
adamcpovey: (a) Please email me a path to one of these RSS images. They clearly matter to you and so should be part of my regression tests.
(b) Could you explain what HRIT means?
(c) My issue with the line is that if HRIT is false, which it is for the file I use in my regression test, it uses the SEVIRI reader exclusively. We believe that that is presenting an angle the wrong way around and so cannot be subset correctly.
(d) The refactor would be a clause looking for subsetting. I'd prefer the SEVIRI reader be corrected, but I can't do that.

adamcpovey:
I've got more changes to upload but want to be absolutely certain of them first.

adamcpovey:
As there are other bugs kicking around this PR, I'm ending it and will start another one once I sort out my code.

PR #29, opened 2020-06-16
Preprocessor updates to surface reflectance

simonrp84:
This PR does two things:
1) Prevent calculation of surface reflectance + BRDF data for night-time pixels. This data is unnecessary (as it's night, there's no solar reflectance) and just takes up processing time.
2) Remove a hack in the cox_munk code that was used to flip the relative azimuth angles. Now that we're fixing the RAA in the satellite readers we no longer need this flip.

### DIFF
--- a/pre_processing/cox_munk.F90
+++ b/pre_processing/cox_munk.F90
@@ -446,10 +446,6 @@ subroutine cox_munk(bands, solza, satza, solaz, totbsc, totabs, relaz, u10, v10,
 
    rsolaz = 0.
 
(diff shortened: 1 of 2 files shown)


adamcpovey:
Shouldn't you also update the comments to represent the new reality of the code in Cox-Munk?

I haven't finished going through the maths yet as I've been taking some time off around meetings. Am like 90% sure this is the right choice.

simonrp84:
I forgot that this was the one bit of code with good comments in it. ;-)
Removed now, thanks.

PR #32, opened 2020-07-27
Subsetting SLSTR and SEVIRI

adamcpovey:
A second attempt at #28 updating the read code of these instruments to allow subsetting of orbits (which I use for processing only the pixels around an AERONET station). This is mostly the contents of my last pull request, with some out-of-bounds errors fixed. I can only hope the SEVIRI decision tree is to Simon's liking. I take on-board the action of remaining vigilant for the needed update of `seviri_util`.

Splitting these off my own fork of ORAC is a god-forsaken mess so I'm trying adding a branch to the main repo, which will be deleted after the PR is resolved.


pre_processing/read_slstr_funcs.F90
```diff
@@ -326,42 +315,106 @@ subroutine slstr_resample_vis_to_tir(inarr, outarr, nx, ny, fill)
...
 
    integer :: x
-   real    :: bdiff, summer
+   real    :: bdiff, total
+#endif
+
+   call get_slstr_gridsize(indir, 'in', nx, ny)
+   call get_slstr_gridsize(indir, 'io', obnx, obny)
+
+#ifdef CONSTANT_OBLIQUE_OFFSET
```
simonrp84, line 346: I am very uncomfortable with this being the default option, the number of cases where `CONSTANT_OBLIQUE_OFFSET` != 548 is high enough to be troubling, which will lead to our aerosol retrieval being wrong (slightly) in those cases. Can we switch this around (or at least discuss in the next ORAC meeting) so that calculating the offset is the default option? It's not a very time-consuming piece of code anyway, so won't have a big performance impact (compared to RTTOV or the full BRDF).

adamcpovey: I didn't do this for efficiency reasons. From an engineering point-of-view, there *should* only be one offset and that number *should* drop out of the L0->L1 processing.

Coastlines look like shit with line-by-line code on and, to my eye, what our search does is remove the recently introduced rejigging of lines to improve collocation. I'll check the quality of the aerosol retrieval between the two modes in some validation I'm doing. Will be impressed if I get it done by Wednesday, though.

simonrp84: Yes, I've not checked this with the latest baseline - so maybe it's unneeded now.
What time period of data did you use for your analysis that determined 548 was the right number? If it's not all post-Jan 2020 then any chance you coudl run something just to see how it looks now?

adamcpovey:
I have now

1. Merged the changes that @danielphilipp made to seviri_util, which removed the need for my fudge when subsetting SEVIRI and got rid of the need for the `if full_disc` statement @simonrp84 understandably didn't like.
2. Changed the standard nadir-oblique alignment from a compilation constant into a driver file parameter. I shall soon see if this is better than Simon's search by default.
3. I added a generic check that azimuths are less than 180deg, as they weren't for SLSTR or SEVIRI. This can be moved inside those specific reader if we'd prefer, but I feel this is an issue liable to recur with future instruments and this is an easy way to catch that.

pre_processing/read_slstr.F90
```diff
@@ -15,6 +16,29 @@ module read_slstr_m
...
+        0.92, & ! Ch10
+        0.93, & ! Ch11
+        0.93, & ! Ch12
+        -1.0, & ! Ch13
+        -1.0, & ! Ch14
+        -1.0, & ! Ch15
+        -1.0, & ! Ch16
+        -1.0, & ! Ch17
+        -1.0 ]  ! Ch18
+
```
simonrp84, line 41: Can we not just set the `-1.0` values to `1.0` and then remove the `if slstr_correction_factor(i) .gt. 0` statement in the `read_slstr_funcs.F90`

pre_processing/read_slstr_funcs.F90
```diff
@@ -100,11 +104,11 @@ subroutine get_slstr_startend(imager_time, fname, ny)
    jd2    = jd2 + dfrac2
 
    ! Compute linear regression slope
-   slo = (jd2 - jd1) / ny
+   slo = (jd2 - jd1) / n_along_track
 
    ! Put correct julian date into each location in the time array
    do j = 1, ny
-      imager_time%time(:,j) = jd1 + (slo * float(j))
+      imager_time%time(:,j) = jd1 + (slo * float(j + starty - 1))
```
simonrp84, line 111: Can you remind me why we need a `-1` here?

simonrp84 (commented):
A couple of comments / questions, otherwise looks OK.

PR #36, opened 2020-09-24
Fixed CPH ice phase uncertainty

danielphilipp:
Fixed the cloud phase (CPH) uncertainty estimates for ice phase pixels.


simonrp84 (approved):

simonrp84:
LGTM, approved. Let's wait for @adamcpovey's opinion and then merge, assuming he's OK with it.

adamcpovey (approved):
It's weird for `norm_diff_cc_th` to be implicitly negative, but you're consistent with the existing weirdness of the NN routines.

danielphilipp:
I guess the problem here is that for ice phase, the slope `m` of the linear function (`m*x+b`) is actually negative but in ORAC  the slope is replaced by `UNC_MAX - UNC_MIN` (divided by `DIFF_1 - DIFF_0` what is actually  1 so it is omitted). This will always be positive.  Maybe this "manually calculated slope" was introduced to avoid negative uncertainties as with the linear fit parameters negative uncertainties can occur for `norm_diff_cc_th` close to 1 but I also think this could have been handled in a more straight foreward way ...

PR #38, opened 2020-12-02
USGS file orientation checking

danielphilipp:
As previously discussed, I had issues with the orientation (lat flipped) of the USGS input file containing LSM, DEM, LUS. To solve this I implemented a fix checking whether the USGS input file is upright (lat starts positive) or flipped (lat starts negative) before calculating the nearest neighbour index.


adamcpovey:
I would have multiplied by minus one inside `nearest_USGS`, but what you've done looks appropriate.

simonrp84:
LGTM!

PR #39, opened 2021-01-21
Added SEVIRI and SLSTR spectral response correction

danielphilipp:
Added SEVIRI and SLSTR spectral response correction coefficients and enabled the use of those in pre_processing/cloud_typing_pavolonis.F90.


RoyGrainger:
Dan can you be explicit about SEVIRI and SLSTR spectral response correction coefficients - I am building LUTs based on the spectral response functions at https://www.nwpsaf.eu/site/software/rttov/download/coefficients/spectral-response-functions/
Are there improved spectral response functions available?

simonrp84 (changes requested):
Hi Daniel, looks good!
I just wanted to check: Do the coefficients in `cloud_typing_pavolonis` also need changing?
https://github.com/ORAC-CC/orac/blob/1efcf2a0982df52588674d1f6b10ee26f3dfb5e7/pre_processing/cloud_typing_pavolonis.F90#L2077
and
https://github.com/ORAC-CC/orac/blob/1efcf2a0982df52588674d1f6b10ee26f3dfb5e7/pre_processing/cloud_typing_pavolonis.F90#L2073

I'm pretty sure that I added these, and I remember having a lot of confusion as to how to derive them (there's a description in the file, but the numbers for some of the other satellites - such as MODIS - don't match the values calculated via the description. In the end I used the RTTOV method instead.

Could you take a look at these numbers and see if they are correct? Maybe we also need different numbers for each of the MSGs, as each has a different spectral response?



I've also added one in-line comment to the code.

common/coefficients_spectral_response_correction.inc
```diff
@@ -297,8 +297,8 @@
   ! a=6: n channels
   ! b=2: slope/offset
   ! c=2: day/night_tw
-  ! d=22: satellite
-  real, DIMENSION(6,2,2,22) :: n19mimic = reshape( &
+  ! d=28: satellite
+  real, DIMENSION(6,2,2,28) :: n19mimic = reshape( &
```
simonrp84, line 301: Could you add a brief docstring (or whatever they're called in FORTRAN world!) describing how these values are calculated? Doesn't need to be in huge detail, just something we can use as a quick reference if we need to know in the future.

danielphilipp:
@simonrp84 Will have a look at the Planck stuff next week. My first impression is that there are changes needed ... But I will ask Martin first as Stephan Stapelberg and Oliver Sus likely implemented the SR correction for the ANN and several sensors, maybe they already spent some thoughts on it.

common/coefficients_spectral_response_correction.inc
danielphilipp: Will do that :)

danielphilipp:
@RoyGrainger This correction is only applied temporarily. It solely affects the neural net cloud masking and cloud typing. The idea was that the neural net for CMA and CPH was trained against NOAA-19 data but as other sensors should also use this neural net, their data get linearily corrected to mimic the spectral response of NOAA-19. Maybe it is important to explicitly say that the data are linearily corrected and not the SRF itself. After the CMA and CPH determination the uncorrected data are used again. Basically all sensors were implemented (already for CCI phase 2) but SEVIRI and SLSTR were not. There is a new ANN for SEVIRI but for the sake of completeness I implemented coefficients for both sensors.

adamcpovey:
tldr; When do we abandon the AVHRR cloud mask and instead map all sensors onto a geostationary sensor, using one of the Daniel's neural nets?

To put in my two cents for posterity (not requesting changes on this PR), there are two ways we can interface different sensors with the cloud masking neural nets:

- Prepare a separate net for each sensor using a common methodology;
- Homogenise all sensors into a common bandpass, which feeds into a single net.

The former can make optimal use of the channels available on each instrument and is easy to understand but requires a lot of work to generate training data for every sensor ever launched and will be inconsistent between them. The latter is theoretically a gargantuan task and can never do better than AVHRR but is consistent.

We appear to be preparing separate flags for SEVIRI, Himawari and AVHRR but historically built ad hoc homogenisation approaches. Sayer published a methodology to do something conceptually similar with surface reflectance data, doi:10.1016/j.rse.2011.02.029. I asked for a Master's student to develop a neural net that could replicate Sayer's work, but didn't get one.

IMHO, I like the homogenisation approach - I always know what I'm getting out of the flag and don't have to worry about cloud fraction meaning different things in the 1980s and 2010s. Newer sensors can have a second flag that focuses on catching thin or high clouds, which is a more targeted task for neural net development. The question is then which sensor forms the baseline:

- AVHRR is the lowest common denominator;
- MODIS is the standard against which everything else is calibrated so there exist match-up datasets;
- SEVIRI has at least daily collocations with everything and observes throughout the day, providing a massive dataset from which to build products;
- Himawari has those features and is a better sensor.

Considering we're actively building neural nets for the geostationary sensors, and want to do more processing of those, I feel it makes sense to map the LEO sensors onto one of the GEO sensors and switch to a new mask (once ready). If a project needs us to be better than that for a specific sensor, we can build a second, individualised neural net that is trained to (un)catch clouds that the default mask misses. Users then have the option of masking by <long-term consistent mask> and/or <local, better mask> depending on the time horizon of their work.

danielphilipp:
The reason why I implemented the coefficients for SLSTR and SEVIRI is that we can do cloud masking and cloud phase determination consistently throughout all sensors and not just those used in CCI. For sure with this approach we are not using the full spectral capability of some sensors as we reduce them to the AVHRR heritage channels but we have consistent masks and phase throughout many sensors/satellites. Whether the kind of "homogenization" we are doing is the best I cannot answer now. Basically, I also support the approach "Homogenise all sensors into a common bandpass, which feeds into a single net". Following this, the idea behind the new SEVIRI ANN is what Adam mentioned in the last paragraph: We wanted to explore the maximum CMA/CPH performance we can get out of SEVIRI using the full spectral information available. I would classify it as an external optional module (as it is not used by default) but I would offer it within the ORAC download so that users also have the opportunity to apply this potentially better mask as the work has already been done.

danielphilipp:
> Hi Daniel, looks good!
> I just wanted to check: Do the coefficients in `cloud_typing_pavolonis` also need changing?
> https://github.com/ORAC-CC/orac/blob/1efcf2a0982df52588674d1f6b10ee26f3dfb5e7/pre_processing/cloud_typing_pavolonis.F90#L2077
> 
> 
> and
> https://github.com/ORAC-CC/orac/blob/1efcf2a0982df52588674d1f6b10ee26f3dfb5e7/pre_processing/cloud_typing_pavolonis.F90#L2073
> 
> I'm pretty sure that I added these, and I remember having a lot of confusion as to how to derive them (there's a description in the file, but the numbers for some of the other satellites - such as MODIS - don't match the values calculated via the description. In the end I used the RTTOV method instead.
> 
> Could you take a look at these numbers and see if they are correct? Maybe we also need different numbers for each of the MSGs, as each has a different spectral response?
> 
> I've also added one in-line comment to the code.

Had a chat with Martin yesterday about the Plank coefficients and he thinks they don't need to be adapted/corrected. Actually there might be different numbers for the different MSGs (even though they shouldn't differ massively) but I would like check it and add changes with a future PR if necessary.

common/coefficients_spectral_response_correction.inc
danielphilipp: Resolved with b86c2c6

simonrp84:
Thanks for the changes, @danielphilipp! I agree that the Planck, coefs can wait for another time.

simonrp84:
I'd like to test this before merging, but am unsure if I have the time in the near future. Is anyone else willing to give this a shot?

danielphilipp:
Could anyone give these changes a test so that I can close this branch ? FYI: I just added the spectral band adjustment coefficients for SEVIRI and SLSTR (produced with the same code as those for MODIS, AATSR, ATSR2 and AVHRR which were used with phase 2 of the Cloud_cci project). It is just a extension, nothing new. Not applying this correction could lead to incorrect CMA and CPH results.

PR #41, opened 2021-02-17
Shift SEVIRI satellite azimuth range to match ORAC definition

danielphilipp:
The SEVIRI reader outputs the satellite azimuth in the range [0,360] whereas ORAC expects the satellite azimuth to be in the range [-180,180]. Thus the main processor sets pixels with satellite azimuth to fill value so that half of the SEVIRI disk is missing. 

Introduced a shift of the satellite azimuth within the reader by 180 degrees to match the ORAC definition.

New satellite azimuth outputted by the main processor looks correct:

satazi_new

Please note that the image is east-west flipped so that the satazi increases counter-clockwise. In the correct orientation it increases clockwise.

### DIFF
--- a/pre_processing/read_seviri.F90
+++ b/pre_processing/read_seviri.F90
@@ -262,6 +262,9 @@ subroutine read_seviri_l1_5(l1_5_file, imager_geolocation, imager_measurements, 
       end where
    end where
 
+   ! Shift satazi range from [0,360] to [-180,180] to be consistent with ORAC
+   imager_angles%satazi = imager_angles%satazi - 180.
+
    deallocate(tmparr)
 
 end subroutine read_seviri_l1_5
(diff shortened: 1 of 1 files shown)


simonrp84:
LGTM!
Can I just check: The image above has North at the bottom? :)

adamcpovey:
Don't you mean to do,
```
    where (imager_angles%satazi .gt. 180.)
        imager_angles%satazi = imager_angles%satazi - 360.
    end where
```
I don't remember us using that number inside ORAC, so it doesn't matter that much, but what you've written would define azimuth as being zero to the south, which is rather atypical.

danielphilipp:
Discovered that correcting the satellite zenith angle in https://github.com/gmcgarragh/seviri_util 3 months ago changed the satellite azimuth orientation. Just corrected the satellite azimuth in `seviri_util`. Satellite azimuth and zenith angle calculation is now consistent with https://celestrak.com/columns/v02n02/. Revised this PR with 3a30019 to output the satellite azimuth in the correct orientation. Correcting `seviri_util` allows to remove an array flip within ORAC.

satazi_final

PR #45, opened 2021-03-10
Pedantic renaming

adamcpovey:
Two large but non-operational commits I did to my own code ages ago and need to deal with before I get to the operational stuff.

1. The use of CamelCase filenames in the main processor code is inconsistent with the remainder of the code, the routine names, and trips me up regularly.
2. The verbose output of `ncdf_read_array` makes up the majority of the code's printing without containing anything overly useful in almost all circumstances. On the rare occasions I do want to know it, the details removed in 2cfa8f394818d2ab21d22cd24ff0e5b7f4287e5f would be useful to me. Thus, I'm punting the whole thing into DEBUG flags. If something's gone wrong enough you want to see the particulars of every single variable read, you're debugging and can recompile to do so.

Does anyone else think these are sensible changes?


derived_products/broadband_fluxes/dependencies.inc
```diff
@@ -13,7 +13,7 @@ $(OBJS)/two_rt_lw.o: $(OBJS)/kinds.o
 $(OBJS)/two_rt_lw_iter.o: $(OBJS)/kinds.o
 $(OBJS)/two_rt_lw_ocastrndm.o: $(OBJS)/kinds.o
 $(OBJS)/two_rt_lw_sel.o: $(OBJS)/kinds.o
-$(OBJS)/two_rt_sw_bs.o: $(OBJS)/kinds.o
+$(OBJS)/two_rt_sw_bs.o: $(OBJS)/kinds.o $(OBJS)/bandsolve.o
```
simonrp84, line 16: Where did this pop up from? Was something broken before or is this for a new feature?

pre_processing/dependencies.inc
```diff
@@ -66,8 +66,9 @@ $(OBJS)/read_avhrr.o: $(OBJS)/calender.o $(OBJS)/channel_structures.o \
         read_avhrr_time.F90 read_avhrr_time_lat_lon_angles.F90
 $(OBJS)/read_camel_emissivity.o: $(OBJS)/preproc_constants.o
 $(OBJS)/read_cimss_emissivity.o: $(OBJS)/preproc_constants.o
-$(OBJS)/read_himawari.o: $(OBJS)/channel_structures.o \
-        $(OBJS)/imager_structures.o $(OBJS)/preproc_constants.o
+$(OBJS)/read_himawari.o: $(OBJS)/calender.o $(OBJS)/channel_structures.o \
+        $(OBJS)/imager_structures.o $(OBJS)/preproc_constants.o \
+        $(OBJS)/solar_position.o
```
simonrp84, line 71: Likewise this one, was something broken when we were missing solar_position?

simonrp84 (commented):
Couple of comments, although I now realise these are probably from your renaming rather than from adding a new thing to the compile.

derived_products/broadband_fluxes/dependencies.inc
adamcpovey: This one is curious. The file invokes the dependency at https://github.com/ORAC-CC/orac/blob/c3cacad2266d9666a502bd29c5a67e9e84d2f75a/derived_products/broadband_fluxes/bugsrad/two_rt_sw_bs.f#L19 but `make depend` skips it. Thus, this is me manually adding it in to prevent parallel make from failing.

pre_processing/dependencies.inc
adamcpovey: This was automatic, seeming because of https://github.com/ORAC-CC/orac/blob/c3cacad2266d9666a502bd29c5a67e9e84d2f75a/pre_processing/read_himawari.F90#L76 and https://github.com/ORAC-CC/orac/blob/c3cacad2266d9666a502bd29c5a67e9e84d2f75a/pre_processing/read_himawari.F90#L181

src/Makefile
```diff
@@ -219,13 +219,13 @@ PARSER_NAME     = read_ctrl
 PRINTER_NAME    = print_ctrl
 PARSER_SOURCE   = $(ORAC_COMMON)/common_constants.F90 \
                   $(ORAC_COMMON)/constants_cloud_typing_pavolonis.F90 \
-                  $(ORAC_COMMON)/orac_indexing.F90 ORACConstants.F90 Ctrl.F90
+                  $(ORAC_COMMON)/orac_indexing.F90 orac_constants.F90 ctrl.F90
 USE_STATEMENTS  = "use Ctrl_m; use ORAC_constants_m"
 PLACE           = $(OBJS)/$(PARSER_NAME)
 PARSER_INC      = -I$(ORAC_COMMON) -I../src/$(OBJ)
 include $(ORAC_COMMON)/Makefile_struct_parser
 $(OBJS)/ReadDriver.o: $(PLACE).f.o
```
danielphilipp, line 227: I had to replace `ReadDriver.o` with `read_driver.o` (`read_driver.o` is in the `$(OBJS)` list above) to compile the code.

src/Makefile
```diff
@@ -219,13 +219,13 @@ PARSER_NAME     = read_ctrl
 PRINTER_NAME    = print_ctrl
 PARSER_SOURCE   = $(ORAC_COMMON)/common_constants.F90 \
                   $(ORAC_COMMON)/constants_cloud_typing_pavolonis.F90 \
-                  $(ORAC_COMMON)/orac_indexing.F90 ORACConstants.F90 Ctrl.F90
+                  $(ORAC_COMMON)/orac_indexing.F90 orac_constants.F90 ctrl.F90
 USE_STATEMENTS  = "use Ctrl_m; use ORAC_constants_m"
```
danielphilipp, line 223: Shouldn't `use Ctrl_m; use ORAC_constants_m` be converted to `use ctrl_m; use orac_constants_m` ? Fortran isn't case sensitive but for the sake of completeness shouldn't it be consistent with the rest of the filenames ?

danielphilipp:
Except for the first comment I had no issues compiling the code on ECMWF HPC. So I'm fine with it when the issue in the first comment is resolved.

src/Makefile
adamcpovey: Yup, that's wrong. Gimme a sec...

src/Makefile
adamcpovey: As you say, Fortran isn't case sensitive so I didn't feel the need to fix this. However, since I'm in the file anyway...

danielphilipp (approved):

Issue #44, opened 2021-03-09
Separate instrument calibration factors from the readers


simonrp84:
Currently, if an instrument requires calibration coefficients to be applied (multiplying or adding a value to radiances in order to correct them) this is done within the individual reader. An example of this is SLSTR.

Instead, we should have a separate preprocessor routine that corrects the radiances *after* the reader has loaded the data. We should also include the coefficients in `setup.F90`. The default being 1 for gain or 0 for offset.

PR #46, opened 2021-03-10
Add instrument-specific bias correction to setup, addressing #44

adamcpovey:
How's that? Move the SLSTR calibration adjustment into `read_imager` in order to be applied to any sensor, should we be so inclined.


simonrp84:
I'm thinking about this again and it might need a more complex solution than we originally thought. This works fine for situations like SLSTR where the corrections are applied to the reflectances, but will cause issues when the corrections have to be applied to radiances (before the rad->ref or rad->bt calculation is performed.
Not really sure what the best solution is, does anyone have some thoughts on this?

adamcpovey:
I was thinking of forward model bias correction rather than calibration, which will be in radiance space. (Don't been asking me to bias correct the aerosol algorithm against AERONET for years.)

Other thoughts appreciated

adamcpovey:
> I'm thinking about this again and it might need a more complex solution than we originally thought. This works fine for situations like SLSTR where the corrections are applied to the reflectances, but will cause issues when the corrections have to be applied to radiances (before the rad->ref or rad->bt calculation is performed.
> Not really sure what the best solution is, does anyone have some thoughts on this?

Looking back at this, when do you expect us to apply such a calibration? I thought the only place we do rad->bt calculations is in the forward modelling of mixed channels radiances.

While I can see that the difference between correcting radiances and reflectances is important, the construction of the code is such that we can *only* deal with radiances within the instrument reader functions. If you're concerned that some instruments will want to adjust radiance and others reflectance, that implies that these corrections go in the readers and are bespoke. In that case, this code is intended for forward model errors and I should probably relabel it a bit.

Thoughts?

PR #47, opened 2021-03-11
Feature ahi hsd subset

andyprata:
I've now implemented the subsetting feature for the Himawari/AHI reader within ORAC. It can be used with the `--limit` flag with calls to `orac.py` consistent with how the SEVIRI subsetting is done.

I am suggesting a merge to the `HSD_PyORAC` branch as this is the branch from where I began making changes to implement this feature.

I will leave it up to @simonrp84 or @adamcpovey to do the merge of `HSD_PyORAC` to `master` as I'm not sure exactly what changes have been made there.

Other minor changes contained within this pull request not directly related to AHI subsetting are:
1. Ignore PyCharm-related files in the .gitignore file
2. Bug fix in reading the sensor name for AHI in definitions.py
3. New a priori/errors for the 'EYJ' particle type. I'm currently experimenting with what I think this values should be.


adamcpovey:
There are *a lot* of commits in there that are later reverted. How much of a pain would it be to to remove those commits from the branch, so we only have the net changes?

At the very least, anything that touches `local_defaults.py` needs to go. As is, no one will be able to pull this without deleting their own copy of that file (which is why it wasn't in the repository in the first place; by all means track it in your own fork but not here).

PR #49, opened 2021-03-11
Update style of the python files to match PEP8

simonrp84:
Currently, the python files contain a lot of nonstandard python code. In this PR I attempt to standardise a bit.
Please check that I haven't broken anything in your workflows!


tools/pyorac/swath.py
```diff
@@ -450,26 +448,26 @@ def set_cldflag(self, size=3, low_res=False, cld_limit=3):
                              mode='same', boundary='symm') / num
             e_x2 = convolve2d(aot_tmp * aot_tmp, kernel,
                               mode='same', boundary='symm') / num
-            stdev = np.ma.masked_invalid(np.sqrt(e_x2 - e_x*e_x))
+            stdev = np.ma.masked_invalid(np.sqrt(e_x2 - e_x * e_x))
             self._cldflag[stdev > 0.1] += CLDFLAG["stdev"]
 
-        def opening_test(data, kernel, limit):
+        def opening_test(data, ikernel, limit):
```
adamcpovey, line 454: Since when is kernel a protected term?
And I use an i prefix to indicate indices/counters. Here I'd prefer to shorten the variable to `kern` if there's a clash

tools/pyorac/arguments.py
```diff
@@ -241,17 +242,17 @@ def args_cc4cl(parser):
     phs = cccl.add_mutually_exclusive_group()
     phs.add_argument('-s', '--settings', type=str, action='append',
                      default=None, help='Parameters for each phase to be '
-                     'processed. Each element is a string listing all the '
-                     'arguments to be applied for that phase. This is processed '
-                     'using the same parser so all arguments listed for the '
-                     'main processor are available.')
+                                        'processed. Each element is a string listing all the '
```
adamcpovey, line 245: Ugh. This spacing is *so inefficient* on a short line length but this will look like trash on my editor.
I shall have to pull a copy and redo all the line breaks. <grumble>

tools/pyorac/arguments.py
```diff
@@ -416,7 +418,7 @@ def check_args_cc4cl(args):
     from pyorac.local_defaults import log_dir, extra_lines, retrieval_settings
 
     # Add extra lines files
-    extra_lines.update({key + '_extra' : val for key, val in args.extra_lines})
+    extra_lines.update({key + '_extra': val for key, val in args.extra_lines})
```
adamcpovey, line 421: I have already done a lot of this line respacing (last time I tried out PyCharm it did a lot of it automatically) but it's combined with changing the case on the `local_defaults` variables. I'll see if I can merge this with that.

tools/pyorac/swath.py
```diff
@@ -582,15 +579,16 @@ def __getitem__(self, name):
...
     # -------------------------------------------------------------------
     # Masking functions
     # -------------------------------------------------------------------
-    def mask_cloud(self, allowing=[]):
+    def mask_cloud(self, allowing=None):
         """Bool array that is True where the aerosol cloud flag is non-zero.
 
         The allowing argument can be used to exempt a list of members of the
         orac.CLDFLAG bitmask."""
+        if allowing is None:
```
adamcpovey, line 590: Probably need a second clause that listifies allowing if you pass a single string.

tools/convj.py
```diff
@@ -6,37 +6,19 @@
 :Website: http://github.com/jiffyclub
```
adamcpovey, line 6: Huh. Hadn't look through these files before. This looks like code I get out of other modules.

tools/pyorac/swath.py
simonrp84: Yeah, it's not. The problem that it was shadowing the same name in the outer scope. I then accidentally changed both to `ikernel`. I've now changed the function variable name to `kern` and left the outer scope variable as `kernel`.

tools/pyorac/arguments.py
simonrp84: Updated!

tools/pyorac/arguments.py
simonrp84: Yep, it's all automatic - i just checked afterwards to make sure it wasn't being stupid.

tools/convj.py
simonrp84: It's a mess. I have no idea what's going on in there but at least it's now a slightly tidier mess.

tools/pyorac/swath.py
simonrp84: True. fixed!

adamcpovey:
Yeah, this seems fine. I'll do the capitalisation of `local_defaults` another time, as that will require people to act so it should be done in a de minimis pull.

PR #50, opened 2021-03-12
Add support for Himawari/AHI subsetting

andyprata:
This pull request adds support for Himawari/AHI subsetting. It can be called from `pyorac` with the limit flag (e.g. `--limit X0 X1 Y0 Y1`) consistent with the SEVIRI subsetting approach.


pre_processing/read_himawari.F90
```diff
@@ -223,7 +270,7 @@ subroutine read_himawari_bin(infile, imager_geolocation, imager_measurements, &
    real(kind=sreal)   :: rye, rhr, rminu
```
adamcpovey, line 270: Just caught an unused variable warning in this routine. What should `tlat, tlon` be doing? They aren't allocated but are flagged for multithreading.

andyprata: I didn't touch these variables in this PR, but good catch. Maybe a question for @simonrp84? Can we delete these?

simonrp84: Those variables are needed for speeding up processing with the PGI compilers, so are necessary.

PR #52, opened 2021-04-06
Improve formatting of Python scripts

adamcpovey:
This commit:
 - Uses block capitals for global variables defined in local_defaults;
 - Uses a global to define the permissions of newly created directories;
 - Moves the defaults for regression.py into check_args_regress();
 - Extends comments in local_defaults.py.example.

The first one has the consequence that everyone would have to change their `local_defaults.py` file to add a few new variables and capitalise a bunch of things. Thus, this sits in a pull request until we feel it's a good time.


simonrp84:
Is it a good time to get this thing merged?

adamcpovey:
I'm going to merge this as some part of it leaked in, so the repository doesn't work without it.

PR #53, opened 2021-04-06
Reader for netCDF look-up tables

adamcpovey:
This isn't exhaustively tested but, since people are using it, I feel we should start working on the code together.

Outstanding issues:
 - The dimension ordering in the tables is different to that in ORAC. As we
   believe the table ordering is optimal, the LUT interpolation code should
   eventually be updated to reshape the arrays. Then, the copying code and
   temp arrays in read_sad_lut_nc() can be removed.
 - The Python scripts aren't designed to deal with the many and varied
   LUT names that Don is currently producing. If we're going to use inhomogenous
   sets going forward (e.g. different version numbers across sensors), we will
   need to reconsider how phase is input into the scripts.
 - The various volcanic ash types should probably be given a class. I propose
   renaming Eyj to Ash.


src/orac.F90
```diff
@@ -328,7 +328,7 @@ subroutine orac(mytask,ntasks,lower_bound,upper_bound,drifile)
    ! parameters and read the SAD values.
    allocate(SAD_Chan(Ctrl%Ind%Ny))
 
-   call Read_SAD(Ctrl, SAD_Chan, SAD_LUT)
+   call Read_SAD(Ctrl, global_atts%platform, SAD_Chan, SAD_LUT)
```
andyprata, line 331: I've reverted this change in my update to this PR (see https://github.com/ORAC-CC/orac/pull/61). In my update I add the netcdf LUT filename to the `Ctrl` struct, so the platform doesn't need to be parsed here.

src/read_sad_lut.F90
```diff
@@ -17,7 +17,8 @@
 ! 2021/03/08, AP: Gather grid dimensions into LUT_Grid_t
```
andyprata, line 17: I've made multiple changes to this file. Details here: https://github.com/ORAC-CC/orac/pull/61

tools/pyorac/definitions.py
```diff
@@ -549,10 +549,11 @@ def sad_dir(self, sad_dirs, inst):
             if "AVHRR" in inst.sensor:
                 fdr_name = join(fdr, inst.sensor.lower() + "-" +
                                 inst.noaa + "_" + self.sad)
+            if len(self.name) == 3:
+                file_name = "_".join((inst.inst, self.name, "RBD", "Ch*.sad"))
             else:
-                fdr_name = join(fdr, inst.sensor.lower() + "_" + self.sad)
-
-            file_name = "_".join((inst.sensor + "*", self.name, "RBD", "Ch*.sad"))
+                file_name = "_".join((inst.platform.lower(), inst.sensor.lower(),
```
andyprata, line 555: I've added a method to the `ParticleType` class that deals with the netcdf filenames. See here: https://github.com/andyprata/orac/blob/5643ce814e441d5c33c468d046b4af625dbfa8d5/tools/pyorac/definitions.py#L568

src/get_measurements.F90
```diff
@@ -174,19 +173,14 @@ subroutine Get_Measurements(Ctrl, SAD_Chan, SPixel, MSI_Data, status)
...
-
       ! Use values read from SAD_Chan files
       do i = 1, SPixel%Ind%Ny
          ii = SPixel%spixel_y_to_ctrl_y_index(i)
          if (SAD_Chan(ii)%Thermal%Flag > 0) then
             if (len_trim(Ctrl%LUTClass) == 3) then
                ! Old LUTs approach (fixed uncertainty)
                ! Note: NEBT is already squared when old LUT is read in.
-               SPixel%Sy(i,i) = SAD_Chan(ii)%Thermal%NEBT
+               SPixel%Sy(i,i) = SAD_Chan(ii)%Thermal%NEBT ** 2
```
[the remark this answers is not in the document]
andyprata: This should not be squared at all. As noted in the comment above, NEBT is already squared when read in with `read_sad_chan.F90` here: https://github.com/ORAC-CC/orac/blob/89b0d42dd1a0689451947691f0b85e07ab332287/src/read_sad_chan.F90#L208
Line 183 should be `SPixel%Sy(i,i) = SAD_Chan(ii)%Thermal%NEBT`.

src/get_measurements.F90
```diff
@@ -162,9 +177,39 @@ subroutine Get_Measurements(Ctrl, SAD_Chan, SPixel, MSI_Data, status)
...
+               ! Convert to variance by squaring uncertainty
+               SPixel%Sy(i,i) = (SAD_Chan(ii)%Thermal%NEBT * &
+                                 (dR_dT0(1) / dR_dTm(1))) ** 2
+            end if
          else
-            SPixel%Sy(i,i) = SAD_Chan(ii)%Solar%NedR
+            if (len_trim(Ctrl%LUTClass) == 3) then
+               ! Old LUTs approach (fixed uncertainty)
+               ! Note: NEDR is squared when old LUT is read in.
+               SPixel%Sy(i,i) = SAD_Chan(ii)%Solar%NEDR ** 2
```
andyprata, line 197: Again, this should not be squared.

andyprata (commented):
Looks good except for squaring those uncertainties in `get_measurements.F90` when old LUTs are used. They should not be squared - fix that and I think we're done.

tools/pyorac/definitions.py
```diff
@@ -536,12 +536,62 @@ class ParticleType:
...
+    version - LUT version
     """
 
-    def __init__(self, name, inv=(), sad="CCI_A70-A79"):
+    def __init__(self, name, inv=(),
+                 sad="CCI_A70-A79",
+                 mb='m',
+                 atmospheric_model='01',
+                 microphysical_model='old',
+                 version='06'):
```
andyprata, line 562: We should set the default to `07` now that Don has updated to version 7.

adamcpovey: I thought only some tables moved to v7?

simonrp84:
Changes now made and merged. Thanks for your work on this, everyone!

PR #57, opened 2021-08-03
Update SLSTR reader to overcome array bounds error.

simonrp84:
The SLSTR reader currently reads outside the array bounds when resampling TIR data to VIS resolution:
```At line 263 of file read_slstr_funcs.F90
Fortran runtime error: Index '0' of dimension 1 of array 'outarr' outside of expected range (1:1500)
```

This fixes the issue by reading from `1:1500` instead of `0:1499`.

### DIFF
--- a/pre_processing/read_slstr_funcs.F90
+++ b/pre_processing/read_slstr_funcs.F90
@@ -257,7 +257,7 @@ subroutine read_slstr_visdata(indir, inband, outarr, imager_angles, sx, sy, &
    call ncdf_close(fid, 'read_slstr_visdata()')
 
    ! Resample the data to the TIR grid size.
-   call slstr_resample_vis_to_tir(data1, outarr(offset:offset+nx-1,:), &
+   call slstr_resample_vis_to_tir(data1, outarr(offset+1:offset+nx,:), &
         nx, ny, sreal_fill_value)
 
    ! Convert from radiances to reflectances


simonrp84:
@adamcpovey @garethethomas Could you take a look at this and check that it produces the results you expect?

pre_processing/read_slstr_funcs.F90
```diff
@@ -257,7 +257,7 @@ subroutine read_slstr_visdata(indir, inband, outarr, imager_angles, sx, sy, &
    call ncdf_close(fid, 'read_slstr_visdata()')
 
    ! Resample the data to the TIR grid size.
-   call slstr_resample_vis_to_tir(data1, outarr(offset:offset+nx-1,:), &
+   call slstr_resample_vis_to_tir(data1, outarr(offset+1:offset+nx,:), &
```
adamcpovey, line 260: This would explain why things moved by one pixel...
Thought I'd checked, but I am extremely bad at my job.

But I don't think your correction is entirely right. The offset [should be at least one](https://github.com/ORAC-CC/orac/blob/8ded5ea978fa69255a9c48e2b4393b29a60c5df6/pre_processing/read_slstr_funcs.F90#L350). The actual problem might be that we [hardcode zero](https://github.com/ORAC-CC/orac/blob/8ded5ea978fa69255a9c48e2b4393b29a60c5df6/pre_processing/read_slstr_main.F90#L242) and [offset-1](https://github.com/ORAC-CC/orac/blob/8ded5ea978fa69255a9c48e2b4393b29a60c5df6/pre_processing/read_slstr_main.F90#L250) into the function calls and I can't remember why. It's been there for at least 5 years, though.

adamcpovey (changes requested):

PR #58, opened 2021-08-12
Set default LUT interpolation method to Bicubic rather than Linear.

andyprata:
Based on discussions at the last ORAC meeting it is suggested that Bicubic interpolation be used by default in ORAC for Look-Up Table interpolation. I've suggested linear interpolation as a faster but less accurate option in the comments of read_driver.F90.


adamcpovey (approved):

simonrp84:
This is actually making the results a bit worse for Himawari - but they're already pretty bad so I think this is OK.

simonrp84 (approved):

PR #59, opened 2021-08-12
Add flag to force nighttime retrieval.

andyprata:
I've added a flag to force ORAC into thinking it's nighttime regardless of whether it is or it isn't. This is useful for running the retrieval during the day with only thermal channels (so far I'm finding good results when applying it to volcanic ash clouds). I force the retrieval into using nighttime cloud indexing logic by setting `SPixel%Illum(:) = INight` in get_indexing.F90 which subsequently triggers the `cloud_indexing_logic_night` routine. One potential issue is that if `SPixel%Illum` is written out to any of the preproc ancillary files (I haven't checked) then the flag values would be output as night when it's actually day (or twilight).

### DIFF
--- a/src/get_indexing.F90
+++ b/src/get_indexing.F90
@@ -112,36 +112,34 @@ subroutine Get_Indexing(Ctrl, SAD_Chan, SPixel, MSI_Data, status)
    ! Initialise Approach and Class bits
    SPixel%variables_retrieved = (Ctrl%Approach-1) + 8*(Ctrl%Class-1)
 
-
-   ! If the views have different illumination conditions then skip this pixel
-   do i_view = 1, Ctrl%Ind%NViews
-      SPixel%illum(i_view) = MSI_Data%illum(SPixel%Loc%X0, SPixel%Loc%Y0, i_view)
-
-      if ((i_view > 1) .and. (.not. Ctrl%all_channels_same_view)) then
-         if (SPixel%illum(i_view - 1) /= SPixel%illum(i_view)) then
-            status = SPixelIndexing
-            return
-         end if
-      end if
-   end do
-
    ! Force nighttime retrieval
    if (Ctrl%force_nighttime_retrieval) then
-      SPixel%Illum(:) = INight
+      SPixel%Illum = INight
+   else
+      ! If the views have different illumination conditions then skip this pixel
+      do i_view = 1, Ctrl%Ind%NViews
+         SPixel%Illum = MSI_Data%illum(SPixel%Loc%X0, SPixel%Loc%Y0, i_view)
+         if ((i_view > 1) .and. (.not. Ctrl%all_channels_same_view)) then
+            if (MSI_Data%Illum(SPixel%Loc%X0, SPixel%Loc%Y0, i_view-1) /= SPixel%Illum) then
+               status = SPixelIndexing
+               return
+            end if
+         end if
+      end do
    end if
 
    ! Find which channels are either not used based on illumination conditions
    ! and/or are flagged as missing.
    is_not_used_or_missing = .false.
 
-   if (SPixel%Illum(1) .eq. IDay) then
+   if (Spixel%Illum .eq. IDay) then
       ! All channels are used
       do i_chan = 1, Ctrl%Ind%Ny
          if (MSI_Data%MSI(SPixel%Loc%X0, SPixel%Loc%Y0, i_chan) == sreal_fill_value) then
             is_not_used_or_missing(i_chan) = .true.
          end if
       end do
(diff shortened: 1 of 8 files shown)


adamcpovey (approved):
It's a simple change, go ahead. I've no problem with the Illum flag representing the retrieval that happened rather than a real-world environment. (We output the solar zenith angle so the user can decide for themself if its daytime.)

However, I feel like there must be some way to tidy up GetIndexing. At the moment there are three rows of `Ctrl%X` to specify the various cloud retrievals which are then sorted through by one of many `logic` routines. There should be a discussion of how we do this, as what goes into the state vector is non-trivial to work out at the moment.

Also, I note that `SPixel%Illum` is of length `nviews` but the indexing only considers the first value. Presumably it would be better to add some logic to combine all views into one value for the retrieval? (i.e. if any(Twilight) then Twilight else if any (Day) then Day else Night)

simonrp84:
Only one question: How does the user enable/disable this? It's as simple as adding `Ctrl%force_nighttime_retrieval=True` to the driver file, isn't it?

andyprata:
Adam: 
Yep I agree. I did think about how to combine the `do_new_night_retrieval` flag with this `force_nighttime_retrieval` flag but didn't want to touch it until discussing with others. I think we can definitely simplify the logic that follows from `case(AppCld1L)` in get_indexing.F90. The solution I chose is the least invasive (i.e. everything else will work as normal unless you switch on this new flag). 

The routine does consider all views at line 116-126 of get_indexing.F90 and if the illumination conditions are different then the pixel is skipped. So I assume if it's not skipped then all views have the same illumination condition and `SPixel%Illum(1)` can be used in the logic that follows. However, this made me realise that in the case of forcing the nighttime retrieval and setting all illumination conditions to night, we can skip this loop with:
```
   ! Force nighttime retrieval
   if (Ctrl%force_nighttime_retrieval) then
      SPixel%Illum(:) = INight
   else
      ! If the views have different illumination conditions then skip this pixel
      do i_view = 1, Ctrl%Ind%NViews
         SPixel%illum(i_view) = MSI_Data%illum(SPixel%Loc%X0, SPixel%Loc%Y0, i_view)
         if ((i_view > 1) .and. (.not. Ctrl%all_channels_same_view)) then
            if (SPixel%illum(i_view - 1) /= SPixel%illum(i_view)) then
               status = SPixelIndexing
               return
            end if
         end if
      end do
   end if
```
If you think this is OK I'll update the pull request.

Simon: Yes that's correct.

adamcpovey:
Considering we literally never use anything except the first element of `SPixel%illum`, I think the better answer is to:

- Turn `SPixel%illum` into a scalar, removing the array allocation and deallocation.
- Keep the code you give above that throws an error for pixels with inconsistent illumination, but it only sets the variable once.

andyprata:
As suggested I've now set `SPixel%Illum` to an integer. Main changes to the PR are:

1. `SPixel%Illum` now defiend as an integer in spixel.F90 and initialised with a value of 0 in alloc_spixel.F90.
2. `SPixel%Illum` array allocation/deallocation removed.
2. `MSI_Data%illum(SPixel%Loc%X0, SPixel%Loc%Y0, i_view-1)` is used to compare current illumination condition to previous illumination condition in the loop of my previous comment (as `SPixel%Illum` is no longer an array that stores all illumination conditions).
3.  Replaced all instances of `SPixel%Illum(1)` with `SPixel%Illum`.

I've tested the changes successfully on Himawari/AHI data with `force_nighttime_retrieval=True` and `force_nighttime_retrieval=False`  for the following configurations:
1. WAT with channels 3, 4, 5, 7, 14, 15
2. ICE with channels 3, 4, 5, 7, 14, 15
3. EYJ with channels 3, 4, 5, 7, 14, 15
4. EYJ with channels 13, 14, 15, 16

I've also tested the changes on some of the new LUTs and the retrieval so far works as expected.

src/get_indexing.F90
```diff
@@ -112,44 +112,42 @@ subroutine Get_Indexing(Ctrl, SAD_Chan, SPixel, MSI_Data, status)
...
+         end if
+      end do
    end if
 
    ! Find which channels are either not used based on illumination conditions
    ! and/or are flagged as missing.
    is_not_used_or_missing = .false.
 
-   if (SPixel%Illum(1) .eq. IDay) then
+   if (Spixel%Illum .eq. IDay) then
```
adamcpovey, line 135: Did you intend to lower-case P for the rest of this file?

andyprata: Ah good catch. No I did not intend to make the 'p' in `SPixel%Illum` lower case. That's a typo. I've reverted all instances of `Spixel%Illum` to `SPixel%Illum`

PR #60, opened 2021-08-18
Update to the aerosol retrieval

adamcpovey:
This pull request does three things:

- Add a new term to the `QCFlag` bitmask indicating when the retrieval has used `check_limits()` at any point during its iteration.
- Removes the nadir angular term from the state vector for `AppAerSw` retrievals.
- Reduce the acceptable number of degrees of freedom for noise to 1.0 as, with the change above, that is the point at which retrievals appear to become unconstrained.

The first is hopefully useful for all retrieval modes, but was added to identify a large number of barely converged retrievals of 0.01 optical depth in the aerosol retrieval over land. 

The second only affects aerosol retrievals over land. I find it reduces the number of iterations needed to converge, ensures that the number of degrees of freedom for noise behaves similarly to other retrieval modes, and improves agreement with AERONET. But it means we have a second arbitrary parameter in the Swansea forward model (the other being gamma).

The third will only affect processing that requires `QCFlag == 0`. Having eliminated the unconstrained variable, I now find that the retrieval behaves erratically when there is an entire degree of freedom for noise. (To be specific, in most circumstances the DoFN increase as the AOD decrease but, once DoFN >= 1, we tend to get AOD == 0.01 and erratic values of AER.)

### DIFF
--- a/src/check_limits.F90
+++ b/src/check_limits.F90
@@ -66,8 +66,10 @@ subroutine Check_Limits(Ctrl, X, SPixel, status)
    do i = 1, SPixel%Nx
       if (X(SPixel%X(i)) > SPixel%XULim(SPixel%X(i))) then
           X(SPixel%X(i)) = SPixel%XULim(SPixel%X(i))
+          status = LimitHitUpper
        else if (X(SPixel%X(i)) < SPixel%XLLim(SPixel%X(i))) then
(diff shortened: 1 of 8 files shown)


adamcpovey:
Reminder that I would quite like second opinions on this, particularly from @garethethomas

PR #61, opened 2021-10-12
Update new LUT netcdf reader and add functionality for varying NEBT/NEFR uncertainty with measurements

andyprata:
This PR essentially fixes a few bugs in the netcdf LUT reader and adds functionality for varying NEBT/NEFR uncertainty with the measurements. The following should also be noted:

1. Homog and Coreg are not provided in the newluts and so are hard-coded when newluts are selected - this is a step toward separating forward model uncertainty and measurement uncertainty. When using the old LUTs they are read in from the SAD files as before.
2. Gains are currently hard-coded (see L177-L181 in `get_measurements.F90`). The gains should be read in from the level 1 headers, but this requires some discussion before implementing. 
3. Don's new approximation for `F0` variation is used when using newluts because `F1` is not supplied in the newluts for solar channels (this is done in `read_msi.F90`).
4. I have not added the ability to select old LUT sad values from new LUT netcdfs, but we can do this for backward compatibility. 
5. Bug in `locate.F90` is not fixed in this PR, but will be taken care of when we merge to master as Adam has already fixed it on master.

### DIFF
--- a/src/ctrl.F90
+++ b/src/ctrl.F90
@@ -102,6 +102,7 @@
 ! 2017/07/05, AP: Add NAll to track the total number of channels. New QC.
 ! 2017/10/04, GM: Add use_ann_phase.
 ! 2021/04/06, AP: New LUT names.
+! 2021/10/12, ATP: Add new LUT netcdf filenames.
 !
 ! Bugs:
 ! None known.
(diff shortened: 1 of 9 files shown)


simonrp84 (commented):
Couple of comments / questions - otherwise looks OK

PR #64, opened 2021-11-12
Add string trimming for HDF-EOS library calls

danielphilipp:
When compiling ORAC and its dependecies with gfortran 4.8.5 on the DWD servers I got an segmentation fault in the `read_mcdc43X.F90` files where strings are passed to external functions in the HDF-EOS library. Trimming those strings with `trim(adjustl())` solved this issue.

I also reintroduced `U10M`, `V10M`, `T2M` variables to  ERA5 variable names in the `read_ecmwf_nc.F90` and `read_ecmwf_wind_nc.F90` as the DWD ERA5 files contain those.


simonrp84:
LGTM!

PR #66, opened 2021-11-18
Fix ncdf

simonrp84:
ORAC reads netcdf files by first applying a fill value (`-999`) and then applying the scale + offset specified in the netcdf attributes.
This is problematic in the case that `-999` is a valid value in the unscaled netCDF data (such as for short int data in the range `-32766` -> `+32767`).

In this PR I correct the netcdf read code so that this problem doesn't arise any more. I think I also have it set up to cover the case that the input netcdf data doesn't contain a fill value but can't actually check this as I don't have any such netcdf files available.

### DIFF
--- a/common/ncdf_read_field.inc
+++ b/common/ncdf_read_field.inc
@@ -38,17 +38,30 @@
 
    ! replace file's fill value and out of range values with our own and apply
    ! scale factor/offset
-   if (.not. fv_flag) fv = fill
-   if (vr_flag) then
-      where (arr.ne.fv .and. arr.lt.vmin) arr = fv
-      where (arr.ne.fv .and. arr.gt.vmax) arr = fv
-   end if
-   if (sf .ne. 1.0 .or. of .ne. 0.0) then
-      where (arr.ne.fv)
-        arr = sf*arr + of
-      elsewhere
-        arr = fill
-      endwhere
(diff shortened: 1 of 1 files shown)


garethethomas (changes requested):
Actually, I'm not sure this is quite right. If there is no fill value defined for the data, then just setting `fv` to `fill` is dangerous, as it could be a valid value.  I think the only bullet proof way of dealing with this is something like:
```
      if (vr_flag) then
            where (arr.ne.fv .and. arr.lt.vmin) arr = fv
            where (arr.ne.fv .and. arr.gt.vmax) arr = fv
            if (sf .ne. 1.0 .or. of .ne. 0.0) then
                where (arr.ne.fv)
                     arr = sf*arr + of
                elsewhere
                     arr = fill
                endwhere
           end if
      else
           if (sf .ne. 1.0 .or. of .ne. 0.0) then
                 where(arr.ge.vmin .and. arr.le.vmax)
                      arr = sf*arr + of
                 elsewhere
                      arr = fill
                 endwhere
      end if
```

simonrp84:
Ok, good point @garethethomas. My new commit should fix this.

adamcpovey (approved):
Fine with me (by eye)

simonrp84:
Thanks for looking, guys.

PR #68, opened 2021-11-23
Updates to ancillary data readers

simonrp84:
The ocean colour and CAMEL readers are only working for old formats of data. This update makes them work with new versions. Ocean colour must now be a recent version (tested with v5.0, may work with v4.0) but camel can handle either V1 or V2 data.


simonrp84:
Good point, @garethethomas - done this now!

PR #69, opened 2021-11-29
Update ORAC main processor to enable customised filenames for output

simonrp84:
This is primarily a kicking-off point for discussion on how to incorprorate custom file naming into the ORAC main processor.
The issue we have is that currently the filename is decided from the input filename (from the preprocessor) and the LUT class. With the new ORAC LUTs we now have multiple LUT types for the same class (for example, various ice habits) and each of these would produce the same output filename - which is undesirable.
A workaround is to save each LUT output to its own directory but that seems very messy.
Here, I add a variable to the main processor that allows the user to set a 'comment' in their driver file that then gets used in the ORAC filenames. For example, when processing the general habit mixture for ICE I can set:
`Ctrl%FID%L2_comment         = AGG` and the output filename will end with `_AGG.primary.nc`.

This gets around the issue but might not be the best solution. Let's discuss what we should do.
Options I can think of.
1) Keep the status quo. Works with all our existing code but presents problems with multiple LUTs.
2) Add an attribute inside each LUT that adds a LUT specific 'comment' to the filename. This is less flexible but also less likely to break other parts of the code.
3) This approach. Very flexible, but could break the python scripts and post-processor.

### DIFF
--- a/pre_processing/read_himawari.F90
+++ b/pre_processing/read_himawari.F90
@@ -316,6 +316,7 @@ subroutine read_himawari_bin(infile, imager_geolocation, imager_measurements, &
                      .false., & ! True = output as VIS channel resolution, False = Output at IR res .
                      global_atts%Satpos_Metadata, & ! Struct to store the satellite position data, for parallax
                      .true., & ! Compute or not the solar angles in the reader.
+                     .true., & ! Use updated (block 6, true) or original (block 5, false) VIS calibration
                      verbose) .ne. 0) then ! Verbosity flag.
       write(*,*) 'ERROR: in read_himawari_read(), calling ' // &
                  'AHI_Main_Read(), filename = ', trim(infile)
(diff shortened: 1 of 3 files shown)


adamcpovey:
What do you think of how Andy dealt with this?

I'd personally prefer ORAC filenames to be deducible from the L1 file and the LUT alone, as it makes it easier to write "did I process everything" scripts.

simonrp84:
Sorry, how did Andy deal with this?

adamcpovey:
https://github.com/ORAC-CC/orac/blob/c8a841198802cf7a14960c02fc6394a8c580b3da/src/read_driver.F90#L453-L454

and

https://github.com/ORAC-CC/orac/blob/c8a841198802cf7a14960c02fc6394a8c580b3da/tools/pyorac/definitions.py#L585-L591

He uses the full name of the file, with various identifying numbers.

PR #70, opened 2021-11-30
Update MODIS Albedo/Parameter code to take quality flags into account.

simonrp84:
Currently the MCD43 code in the pre processor doesn't take into account the quality flags that accompany the data. Indeed, we don't even read the QA flags.

This leads to issues when the MODIS BRDF cannot be inverted in the upstream code, as it will then fall back to a magnitude inversion or - in particularly bad cases - fill values. In this PR I update ORAC to read the QA flag and apply filtering if requested by the user. By default, no filtering is done (i.e: Behaviour remains same as now).
Users can set the desired QA threshold by passing `MCD43_MAX_QAFLAG=n` via the preprocessor driver file, where `n` is the desired maximum QA value.
QA values are defined as follows:
```
 - 0: best quality, 100% with full inversions
 - 1: good quality, 75% or more with best full inversions and 90% with full inversions
 - 2: relative good quality, 75% or more with full inversions
 - 3: mixed, 75% or less full inversions and 25% or less fill values
 - 4: all magnitude iversions or 50% or less fill values
 - 5: 50% or more fill values
 - 255: fill value
```

Setting `MCD43_MAX_QAFLAG` to `3` for example, will allow pixels with QA values of 0, 1, 2, or 3 to pass while those with values of 4, 5 or 255 will be set to the fill value.

Below is an example of how this can be used to filter bad quality retrievals, in this case cloud contamination in the BRDF magnitude inversion:

### DIFF
--- a/pre_processing/get_surface_reflectance.F90
+++ b/pre_processing/get_surface_reflectance.F90
@@ -115,6 +115,24 @@
 !
 ! Bugs:
 ! None known.
+!
+!
+! Notes:
+! The `mcd43_maxqa` variable allows the user to select the desired 'worst-case'
+! QA flag for the MODIS BRDF input data. Any data with a QA flag *greater* than
+! this value will be filtered out and replaced with a fill value.
+! The default value for `mcd43_maxqa` is 5, which means all input data will be
+! used.
+! MODIS QA values are:
+! 0: best quality, 100% with full inversions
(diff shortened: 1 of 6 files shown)


adamcpovey:
This seems very sensible to me, though I haven't actually run it.

PR #71, opened 2022-01-27
Allow CTP to be set using an auxiliary file

garethethomas:
These changes allow the Cloud-Top pressure first-guess and/or a priori to the set using the "AUX" method (which already exists for setting surface temperature and reflectance). To use this functionality, two things are needed:

1. A CTP NetCDF AUX file containing arrays "ctp" and "ctp_var", specifying the FG/AP values and AP _variance_ estimates respectively.
2. `Ctrl%AP(IPc,:)` and/or `Ctrl%FG(IPc,:)`  need to be set to `SelmAux` in the ORAC processor driver file.

### DIFF
--- a/src/ctrl.F90
+++ b/src/ctrl.F90
@@ -103,6 +103,7 @@
 ! 2017/10/04, GM: Add use_ann_phase.
 ! 2021/04/06, AP: New LUT names.
 ! 2021/10/12, ATP: Add new LUT netcdf filenames.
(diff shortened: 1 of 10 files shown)


adamcpovey (approved):
I probably would have tried to be more general than this, as we feasibly might want to manually set a couple fields, but this is fine for now.

Two comments are potential improvements rather than necessary.

PR #77, opened 2022-07-12
Update SLSTR vicarious calibration adjustment coefficients.

simonrp84:
The SLSTR reader was using out-of-date calibration coefficients for the VIS channels. This PR updates the coefs to the latest version available from the Sen-3 MPC.

### DIFF
--- a/pre_processing/read_slstr.F90
+++ b/pre_processing/read_slstr.F90
@@ -18,23 +18,25 @@ module read_slstr_m
 
    ! The calibration of SLSTR is under ongoing revisions, such that the
    ! application of empirical correction factors is currently required.
-   ! These are taken from the SLSTR User Guide and/or personal communications.
+   ! These are taken from personal communications and the SLSTR Vicarious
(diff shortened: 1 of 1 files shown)


adamcpovey:
The code obviously complies and runs just fine so technically this PR is acceptable.

I ran this and the current repository code over a week of AERONET collocations of Sentinel 3A at the start of 2022. This scatterplot compares aerosol optical depth at 550nm to AERONET, with the orange dots denoting the values after this change and the other end of the blue line denoting the value from the base code. The left plot averages all converged retrievals within 25 km of the site while the right only considers points for which `QCFlag == 0`
PR77

My read of that is that these changes tend to decrease aerosol optical depth. Looking through the individual files, it looks more like AOD over ocean is slightly increased while that over land more substantially decreases.
download3

There are a few cases where the number of converged retrievals increases substantially, which I would usually take as a good sign, but it mostly changes only slightly.

download4

From this minimal check, I'd say these calibrations are fine. I've also got cloud retrievals for these fields but nothing to validate against. COT is down a little while CTH up a little.

adamcpovey (approved):

PR #79, opened 2022-11-22
Simplify Rayleigh scattering handling and fix RTTOV missing vector error

danielphilipp:
As discussed I implemented some changes to rttov_driver.F90:

1. Add `calcrefl` and `reflectance` arrays to ORAC and pass them to RTTOV as they are requested by RTTOV when `opts % rt_ir % addsolar = .true.`. If not passig them, the error is "calcrefl and reflectance parameters required" (rttov13/src/main/rttov_direct.F90 L318). Some compilers seem to implicitly define those arrays and there is no error thrown but with others there is. To solve this explicitly define and pass them to RTTOV. Set `reflectance` to 0. as the surface reflectance is explicitly handled by the foreward model. Set `calcrefl` to false so that we can provide our own reflectance.
2. Yet RTTOV calculates the Rayleigh scattering what is then removed in ORAC. To simplify that switch off the Rayleigh calculations in RTTOV so that they do not need to be removed.

### DIFF
--- a/pre_processing/rttov_driver.F90
+++ b/pre_processing/rttov_driver.F90
@@ -443,11 +443,10 @@ subroutine rttov_driver(coef_path, emiss_path, granule, preproc_dims, &
    ! regression limits (NB by doing this the extrapolated values outside
    ! the regression limits will be reset to the limits: it will not result
    ! in unphysical extrapolated profile values being used)
-   opts % rt_all % use_q2m                = .false. ! Do not use surface humidity
-   opts % rt_all % addrefrac              = .true.  ! Include refraction in path calc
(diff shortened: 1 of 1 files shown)


simonrp84:
To be honest I think we should keep this pull request. It's clearly the correct choice to have Rayleigh disabled within RTTOV and if that causes problems in ORAC then it's within ORAC that they should be fixed, the current 'add-then-remove' method makes me very uneasy...

adamcpovey:
I concur - 'add then remove' was done out of necessity. Now RTTOV lets us turn Rayleigh off, we should do that. It's on my list of things to sort out properly in the new year.

PR #82, opened 2023-06-02
Belated merge of my branch with main code branch

garethethomas:
Quite a few minor changes - hopefully not _undoing_ tidying that Adam has done.

Commit comments should give a reasonable idea of the specific changes implemented. Once accepted (or rejected, fixed, resubmitted and accepted!) I have a few other more self-contained changes to feed back to the main fork.


pre_processing/orac_preproc.F90
```diff
@@ -486,6 +490,9 @@ subroutine orac_preproc(mytask, ntasks, lower_bound, upper_bound, driver_path_fi
    nargs = command_argument_count()
 #endif
 
+   ! Minimal output to notify start and end of programme
```
adamcpovey, line 470: Could this have `if (verbose)` at the front so we maintain silent operation by default?

pre_processing/orac_preproc.F90
```diff
@@ -1405,6 +1412,10 @@ subroutine orac_preproc(mytask, ntasks, lower_bound, upper_bound, driver_path_fi
    ! deallocate optional arguments
    if (associated(preproc_opts%channel_ids)) deallocate(preproc_opts%channel_ids)
 
+   ! Notify of sucessful completion
```
adamcpovey, line 1182: As above

pre_processing/read_aatsr_dimensions.F90
```diff
@@ -80,7 +83,7 @@ end subroutine get_aatsr_dimension
    logical,                        intent(in)  :: verbose
 
    character(len=path_length,kind=c_char) :: l1b_file_C
-   integer(c_short)                       :: tmp_dynght, half_orbit
+   integer(c_short)                       :: tmp_dynght, half_orbit, c_dynght
```
adamcpovey, line 86: I don't think you needed to make an extra variable here, as `tmp_dynght` is only passed to C routines or considered in if statements.

adamcpovey, line 86: I don't think you needed to make an extra variable here, as tmp_dynght is only passed to C routines or considered in if statements.

pre_processing/read_aatsr_orbit.c
```diff
@@ -11,6 +11,7 @@
       and 512.
    2016/10/26, GT: Fixed a bug whereby measurement time was always being read
       from first row of orbit regardless of starty value.
+   2020/05/11, GT: Fixed bug in selection of day/night data in get_aatsr_dimension.
```
adamcpovey, line 14: Is the "bug" just that you wanted to select different orbit sections?

adamcpovey (changes requested):
Simple enough change and quite useful

pre_processing/orac_preproc.F90
garethethomas: I put this in (and the finishing statement) precisely because completely silent operation can be problematic (especially if running in a batch environment), as it is sometimes not entirely obvious if the code actually ran!

pre_processing/read_aatsr_dimensions.F90
garethethomas: I just wanted to avoid any mismatch between integer(kind=sint) and integer(c_short). Can we be sure that they will always be the same?

pre_processing/read_slstr_funcs.F90
```diff
@@ -229,13 +233,31 @@ subroutine read_slstr_visdata(indir, inband, outarr, imager_angles, &
...
+   else
+      nyr = ny*2
+   end if
+   if (nx1 .gt. nx*2) then
+      nxr = nx1
+   else
+      nxr = nx*2
+   end if
+
+   allocate(data1(nxr,nyr))
```
adamcpovey, line 254: I don't think this bit does anything unless there can be more than 1 extra row/column, the possible states are `ny1 == ny*2` and `ny1 == ny*2+1`, which means `nyr == ny1` in both cases.

pre_processing/read_slstr_funcs.F90
```diff
@@ -191,8 +191,11 @@ subroutine read_slstr_tirdata(indir, inband, outarr, sx, sy)
 
 end subroutine read_slstr_tirdata
 
-! Read the nadir-view visible grid data from SLSTR. Need to update with oblique view
-! This assumes that VIS data is twice resolution of TIR (should be correct, 0.5km)
+! Read the nadir-view visible grid data from SLSTR. Need to update with
```
adamcpovey, line 194: I'm guessing the odd-row-numbers happens because of the rectification done at L0 to make coastlines look realistic in the rectangle. This makes me concerned if our method for accumulating visible radiance is appropriate, given we're working on data that's got cosmetic fill pixels and georectification in it. We can't reasonably do this at L0, but should we worry about orphaned pixels or rows that have been shifted a lot relative to their neighbours?

pre_processing/read_slstr_funcs.F90
```diff
@@ -250,7 +272,10 @@ subroutine read_slstr_visdata(indir, inband, outarr, imager_angles, &
    call ncdf_close(fid, 'read_slstr_visdata()')
 
    ! Resample the data to the TIR grid size.
-   call slstr_resample_vis_to_tir(data1, outarr, nx, ny, sreal_fill_value)
+   ! Ensure that the array passed to the resampling routine is the
+   ! expected size
+   call slstr_resample_vis_to_tir(data1(1:nx*2,1:ny*2), outarr, &
```
adamcpovey, line 277: I'm confused now. In theory, `read_ncdf_field` determines the size of the field to read from the array we give it for output. If the actual data is one bigger than what we expected, the read should go off without a hitch.

Could you send me a filename on JASMIN that crashes here, so I can work out what I did wrong?

adamcpovey (changes requested):
This is mostly fine, with a few housekeeping corrections requested.

The major issue is in the fix for odd-row-number SLSTR files, which will break subsetting. Could you send me a filename on JASMIN that crashes here, so I can work out what I did wrong? The code should have been able to deal with "off by one" errors.

adamcpovey:
I tried to make comments on each commit, but that isn't how GitHub works so sorry for the three bits. I'll pull the code and try compiling it once you get me an SLSTR filename.

pre_processing/read_aatsr_dimensions.F90
adamcpovey: Well that's already happening at https://github.com/ORAC-CC/orac/blob/2111284a33fe44ea6ae26d38def159c84682e968/pre_processing/read_aatsr_dimensions.F90#L102

You could explicitly cast to `c_short` on that line and do comparisons against the original variable

pre_processing/read_aatsr_orbit.c
adamcpovey: Right

pre_processing/read_aatsr_dimensions.F90
garethethomas: It's quite possible I don't really understand how Fortran does the type casting, but I was working on the assumption that you can move the value between numeric types using the `=` operator, but should always ensure the type of variables passed in and out of sub-routines matches with what is defined in the sub-routine?
`day_night` is passed in as a Fortran `integer(kind=sint)` but is then passed on to the C function as a `integer(c_short)` . Could passing `day_night` directly to the C function not be broken if the definition of `sint` changed (or indeed, if the definition of `c_short` was different on some future system?

pre_processing/compute_geopot_coordinate.F90
```diff
@@ -46,49 +48,52 @@ subroutine compute_geopot_coordinate(preproc_prtm, preproc_dims, ecmwf)
    ! upper boundaries (wrt height))
    do ij=preproc_dims%min_lat,preproc_dims%max_lat
       do ii=preproc_dims%min_lon,preproc_dims%max_lon
-         ! this is the lowest level=surface, it also has the surface pressure
-         preproc_prtm%phi_lev(ii,ij,ecmwf%kdim+1)=preproc_prtm%geopot(ii,ij)
-         sp=exp(preproc_prtm%lnsp(ii,ij))
+         ! Check to see we have data for this particular pixel
+         if (preproc_prtm%lnsp(ii,ij) .ne. sreal_fill_value) then
```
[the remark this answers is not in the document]
simonrp84: Yes, space pixels will have a fill value for lnsp.

pre_processing/read_avhrr_time_lat_lon_angles.F90
```diff
@@ -87,16 +88,24 @@ subroutine read_avhrr_time_lat_lon_angles(path_to_geo_file, imager_geolocation,
         imager_time, refjulianday)
 
    !read latitude
-   call read_avhrr_lat_lon(geo_id, "where/lat", "data", "where/lat/what", &
-        imager_geolocation%startx, imager_geolocation%endx, &
-        imager_geolocation%starty, imager_geolocation%endy, temp)
-   imager_geolocation%latitude = temp
+   call read_avhrr_lat_lon(geo_id,"where/lat","data","where/lat/what", &
```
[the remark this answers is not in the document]
garethethomas: Done

pre_processing/read_slstr_funcs.F90
garethethomas: Ok, I've looked a bit further into this and it looks like the problem is that some frames, which appear to always be at the  have one extra row in both IR and VIS grids. Therefore, the vis channel has one _less_ row than the code is expecting, which would cause your code to fall over.
I agree with you, having gone through the whole ncdf reading nest of routines, that if there was an extra row in the visible and not the IR, it would read the data just fine.
I've attached a list of files which suffer from this problem (at least most of them do).
[missing_l2_slstrb_20190101_to_20191231-20230531.txt](https://github.com/ORAC-CC/orac/files/11663499/missing_l2_slstrb_20190101_to_20191231-20230531.txt)

pre_processing/read_slstr_funcs.F90
garethethomas: You are right, it is a bit worrying. I'm pretty sure this is not meant to happen and I don't think it is documented, so who knows what the best thing to do is!?
My guess would be that we're unlikely to be more than 1 out on the VIS pixel grid, which is probably better than the co-registration between the IR and VIS grid in the L1B data anyway.

pre_processing/read_slstr_funcs.F90
```diff
@@ -229,13 +233,31 @@ subroutine read_slstr_visdata(indir, inband, outarr, imager_angles, &
...
+   end if
+
+   allocate(data1(nxr,nyr))
 
    data1(:,:) = sreal_fill_value
 
    ! Open the netcdf file
-   call ncdf_open(fid, filename, 'read_slstr_visdata()')
-   call ncdf_read_array(fid, bandname, data1, start=[sx*2-1, sy*2-1])
+   call ncdf_read_array(fid, bandname, data1(1:nx1,1:ny1), &
```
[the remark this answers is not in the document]
garethethomas: Ok, I think I've come up with the workable solution. As mentioned below, although these frames have one extra row of VIS data, they also have one extra row of IR data. Therefore, the problem isn't that the VIS data has one more row  than the code is expecting, but that they have one _LESS_.
I've changed my code so that it deals with this specific issue and allows for exactly one fewer than nx*2 or ny*2 columns or rows in the VIS grid. Any more than that, and it will still fall over. It should also handle the sx and sy parameters.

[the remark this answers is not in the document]
adamcpovey: Cool. Push it up and I'll give it a try

derived_products/broadband_fluxes/process_broadband_fluxes.F90
```diff
@@ -555,9 +555,14 @@ subroutine process_broadband_fluxes(Fprimary, FPRTM, FALB, FTSI, fname,&
    !----------------------------------------------------------------------------
 #ifndef WRAPPER
    index1 = index(Fprimary,'_', back=.true.)
+   ! ORAC filenames without orbit number included
    cyear = trim(adjustl(Fprimary(index1-12:index1-9)))
    cmonth = trim(adjustl(Fprimary(index1-8:index1-7)))
    cday = trim(adjustl(Fprimary(index1-6:index1-4)))
+   ! ORAC filenames with orbit number included (e.g. SLSTR)
```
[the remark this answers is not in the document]
garethethomas: This is a "temporary" hack - the way Matt originally hard coded the extraction of the date from the primary file name breaks if that filename includes an orbit number. It's ugly, but a obvious and easy solution to the problem didn't jump out at me.

pre_processing/read_slstr_funcs.F90
```diff
@@ -235,28 +238,28 @@ subroutine read_slstr_visdata(indir, inband, outarr, imager_angles, &
...
-   if (nx1 .gt. nx*2) then
-      nxr = nx1
+   if (nx1 .lt. nx*2 + sx*2 - 1) then
+      nxr = nx*2 - 1
    else
       nxr = nx*2
    end if
 
-   allocate(data1(nxr,nyr))
+   allocate(data1(nx*2,nyr*2))
```
adamcpovey, line 258: That can't be the correct set of letters, since it mixes one with an r with one without an r.

derived_products/broadband_fluxes/process_broadband_fluxes.F90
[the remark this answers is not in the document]
adamcpovey: That's fine, but at the moment we have to recompile the code to switch between sensors. Why not stick it in an `if (len_trim(Fprimary) == NUMBER) then` clause?

adamcpovey (changes requested):
Closer...

pre_processing/read_slstr_funcs.F90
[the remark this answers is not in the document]
garethethomas: It should be there...

pre_processing/orac_preproc.F90
```diff
@@ -1291,6 +1297,8 @@ subroutine orac_preproc(mytask, ntasks, lower_bound, upper_bound, driver_path_fi
 
       ! perform RTTOV calculations
       if (verbose) write(*,*) 'Perform RTTOV calculations'
+
```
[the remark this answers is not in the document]
garethethomas: I'm not sure which code you're referring to!

pre_processing/setup.F90
```diff
@@ -1428,16 +1432,21 @@ subroutine setup_slstr(args, source_attributes, channel_ids_user, &
...
    read(args%chour, '(I2)') args%hour
    read(args%cminute, '(I2)') args%minute
-   read(args%csecond, '(I2)') args%second
-   if (args%second .ge. 30) then
+   ! SLSTR file times are rounded to the nearest second, whereas the attributes
+   ! are specified to a micro-second.
+   read(args%csecond(1:len_trim(args%csecond)), '(F9.6)') fsecond
+   args%second = NINT(fsecond)
+   ! Deal with the case where rounding the second carries up to a new minute
+   if (args%second .eq. 60) then
```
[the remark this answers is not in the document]
garethethomas: Not sure what you mean.
Previously the code rounded the _minute_ if the second field (which was read as an integer - presumably this would truncate the fractional second, not round it) was `arg%second .ge. 30`. I'll admit that I'm not 100% whether the time-stamp in a L1B filename is _rounded_ to the nearest second or _rounded down_ to the previous second. I suspect the former though.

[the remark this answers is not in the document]
adamcpovey: This no longer rounds to the nearest minute as the outermost if statement `if second == 60`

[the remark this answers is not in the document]
adamcpovey: Off the top of my head, we want to read the seconds as a float, then round the seconds to the nearest whole number, then do the previous code block to round to the nearest minute. But that probably repeats a block of code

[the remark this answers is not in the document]
garethethomas: Exactly - it rounds to the nearest second with the `NINT` statement, and then alters the minutes (and hours and days if necessary) if seconds ends up being 60.
I'm not quite sure what your original question was (?)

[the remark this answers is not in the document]
adamcpovey: Yes, but I thought the problem was that we wanted to round to the nearest minute and we were using the wrong value of seconds? If the problem was actually that SLSTR doesn't round to the nearest minute then you're fine

pre_processing/read_slstr_funcs.F90
garethethomas: Fixed

pre_processing/read_slstr_funcs.F90
garethethomas: Fixed

adamcpovey (approved):

PR #84, opened 2023-06-16
Use WMO platform naming, but strip hyphens out of ORAC filenames

adamcpovey:
_**This will break local scripts as it changes output file names and some driver file arguments. WE ALL NEED TO TEST THIS ONE!**_

This attempts to clean up instrument naming in ORAC. In theory, the pre and main processors will now use the same formatting for sensor/platform names while platform names will follow the WMO standard whereby there's usually a hyphen before any number. Clauses dealing with this are removed from netcdf_output_create_file.F90, read_lwrtm.F90, read_swrtm.F90 and are instead in setup.F90 when parsing input files or in sad_chan.F90 when determining the name of the text LUTs. Names for NCDF LUTs are worked out within the Python script and mostly follows the WMO standard.

An addition is made to preparation.F90 to remove hyphens from ORAC filenames as these are a protected character in CCI filenames. These are put back in by pyorac.FileName.

Theoretically, the main processor no longer requires the Ctrl%InstName
argument as it could be read from the configuration file.

### DIFF
--- a/pre_processing/cloud_typing_pavolonis.F90
+++ b/pre_processing/cloud_typing_pavolonis.F90
@@ -1960,77 +1960,55 @@ function plank_inv(input_platform, T)
 
    ! select appropriate row of coefficient values
    select case (input_platform)
-   case ("noaa5")
+   case ("NOAA-5")
       index = 1
-   case ("noaa6")
+   case ("NOAA-6")
       index = 2
-   case ("noaa7")
+   case ("NOAA-7")
       index = 3
-   case ("noaa8")
+   case ("NOAA-8")
       index = 4
-   case ("noaa9")
+   case ("NOAA-9")
       index = 5
-   case ("noaa10")
+   case ("NOAA-10")
       index = 6
-   case ("noaa11")
+   case ("NOAA-11")
       index = 7
-   case ("noaa12")
+   case ("NOAA-12")
       index = 8
-   case ("noaa14")
+   case ("NOAA-14")
       index = 9
-   case ("noaa15")
+   case ("NOAA-15")
       index = 10
-   case ("noaa16")
+   case ("NOAA-16")
       index = 11
-   case ("noaa17")
+   case ("NOAA-17")
       index = 12
-   case ("noaa18")
+   case ("NOAA-18")
       index = 13
-   case ("noaa19")
+   case ("NOAA-19")
       index = 14
-   case ("metop01")
+   case ("Metop-B")
       index = 15
-   case ("metopb")
-      index = 15
-   case ("metop02")
-      index = 16
-   case ("metopa")
+   case ("Metop-A")
       index = 16
    case ("TERRA")
       index = 17
    case ("AQUA")
       index = 18
-   case ("Envisat")
+   case ("Envisat", "ERS2")
       index = 19
-   case ("ERS2")
-      index = 19
-   case ("MSG1")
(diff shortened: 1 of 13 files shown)


garethethomas:
Sensible and probably necessary change. I've tested with SLSTR and it works just fine, but note that I didn't test using your Python scripts (only a very minor modification needed for my own scripts to make it work).
I can't test with other instruments so easily, but I can try if needed.

garethethomas (approved):
See my previous comment...

pre_processing/cloud_typing_pavolonis.F90
```diff
@@ -1960,77 +1960,55 @@ function plank_inv(input_platform, T)
...
-      index = 20
-   case ("Himawari-8")
-      index = 21
-   case ("Himawari-9")
+   case ("Himawari-8", "Himawari-9")
       index = 21
-   case ("GOES-16")
-      index = 22
-   case ("GOES-17")
+   case ("GOES-16", "GOES-17")
```
simonrp84, line 2005: Could you also add "GOES-18" here?

pre_processing/cloud_typing_pavolonis.F90
```diff
@@ -1960,77 +1960,55 @@ function plank_inv(input_platform, T)
...
       index = 22
-   case ("SuomiNPP")
+   case ("Suomi-NPP", "NOAA-20")
```
simonrp84, line 2007: Likewise, could you add "NOAA-21" to this one.

pre_processing/setup.F90
```diff
@@ -840,8 +849,16 @@ subroutine setup_avhrr(args, channel_ids_user, channel_info, verbose)
...
       str1 = str1(1:j-1)
-      str2 = str2(j+1:)
-      args%platform = trim(adjustl(str2))
+      str2 = adjustl(str2(j+1:))
+      if (str2(1:4) == 'noaa') then
+         args%platform = 'NOAA-' // trim(str2(5:))
+      else if (str2(1:6) == 'metopa' .or. str2(1:7) == 'metop02') then
+         args%platform = 'Metop-A'
+      else if (str2(1:6) == 'metopb' .or. str2(1:7) == 'metop01') then
+         args%platform = 'Metop-B'
```
simonrp84, line 858: Could you add MetOp-C to this?
```
      else if (str2(1:6) == 'metopc' .or. str2(1:7) == 'metop03') then
         args%platform = 'Metop-C'
```

simonrp84 (commented):
The PR looks fine, just a couple of minor comments to bring ORAC up-to-date with the more recent missions.

pre_processing/cloud_typing_pavolonis.F90
adamcpovey: Does it have largely the same spectral response functions? These are subscripting coefficients to make platform X look like the NOAA that goes into the neural network.

If not, additional coefficients will need to be generated.

pre_processing/cloud_typing_pavolonis.F90
adamcpovey: Ditto

simonrp84: Likewise.

pre_processing/setup.F90
adamcpovey: Any idea which of A or B this one looks like?

PR #85, opened 2023-06-22
Add Python interfaces for LUTs and a forward model implementation

adamcpovey:
This adds,
- pyorac.lut, Python containers for both text and NCDF look-up tables;
- pyorac.fm, a Python implementation of the ORAC forward model;
- pyorac.util.bound_grid and bilinear_coefficients, which are used by the forward model routines for interpolating LUTs;
- two Jupyter notebooks demonstrating the use of the above.

I'm not done with this yet, but enough of it works I felt I should throw the code out there for people to look at. Can anyone actually work out how to run the examples? Would they be better as PDFs or hosted somewhere else where I can embed the files alongside? Do any of these things look useful?


adamcpovey:
@garethethomas considering the two people I asked for opinions on this don't work for us any more, who should I throw it to? Elisa or Dora seem the most likely to actually want a Python forward model.

simonrp84:
I never looked at this as I didn't have a need for a forward model outside ORAC.

That said, I've just taken a quick look at the code and it looks OK....aside from various linting issues. I strongly suggest running this through pylint or similar to highlight issues. There's a few places where uninitialised values are used, or class variables aren't referred to with `self.` etc. VSCode or PyCharm will be able ot fix almost all of these automagically.

PR #86, opened 2023-06-23
CEDA/BADC ERA5 data handling

garethethomas:
Contains two main things:
1) A bug fix to the call to the rttov_alloc_transmission subroutine in rttov_driver.f90.
2) Improvements to how the ERA5 data record provided by CEDA/BADC is handled (in set_ecmwf.f90):
   - Temporal spacing for this data is now set to 1-hourly (rather than the default 6-hourly).
   - Code will automatically check for the so-called "provisional" ERA5 data files, if the standard files aren't found (needed if using data newer than 1-2 months).


pre_processing/rttov_driver.F90
```diff
@@ -929,7 +932,7 @@ subroutine rttov_driver(coef_path, emiss_path, granule, preproc_dims, &
          call rttov_deallocate_emis_atlas(emis_atlas)
          call rttov_alloc_traj(stat, 1, nchan, opts, nlevels, coefs, DEALLOC, &
               traj)
-         call rttov_alloc_transmission(stat, transmission, nlevels, nevals, &
+         call rttov_alloc_transmission(stat, transmission, nlevels, nchan, &
               DEALLOC)
          call rttov_alloc_rad(stat, nevals, radiance, nlevels, DEALLOC, &
```
adamcpovey, line 937: As best I can tell, `nevals` appears three times in this code: (1) in your comment, (2) when it's declared, and (3) in `rttov_alloc_rad`. I can't see it being assigned a value, so isn't it also wrong in that call?

pre_processing/rttov_driver.F90
```diff
@@ -929,7 +932,7 @@ subroutine rttov_driver(coef_path, emiss_path, granule, preproc_dims, &
          call rttov_deallocate_emis_atlas(emis_atlas)
          call rttov_alloc_traj(stat, 1, nchan, opts, nlevels, coefs, DEALLOC, &
               traj)
-         call rttov_alloc_transmission(stat, transmission, nlevels, nevals, &
+         call rttov_alloc_transmission(stat, transmission, nlevels, nchan, &
```
adamcpovey, line 935: This also needs to be done in `rttov_driver_gfs.F90`, which Simon added for reasons I have never bothered to question despite it repeating a fuckton of code.

pre_processing/set_ecmwf.F90
```diff
@@ -94,6 +98,10 @@ subroutine set_ecmwf(granule, opts, nwp_flag, imager_geolocation, &
    character     :: cera_year*4, cera_month*2, cera_day*2, cera_hour*2
    character     :: cera_year2*4, cera_month2*2, cera_day2*2, cera_hour2*2
 
+   ! If we're reading BADC ERA5 data, we default to one file every hour,
+   ! rather than 6-hours 
+   if (nwp_flag .eq. 2) opts%nwp_time_factor = 1.0
```
adamcpovey, line 103: I don't think this is the right place for this code. This factor is set by the driver file, and no one would expect that to be overridden in some other routine. Either,

- This should be removed from the driver file and set_ecmwf should contain a case statement that translates `nwp_flag` into an appropriate flag. I assume we didn't do that in the first place because the NWP files might have any spacing if they're downloaded by the user rather than in a repository.
- Your change should be [near here](https://github.com/ORAC-CC/orac/blob/cfe81c518c70f2f552dc420854309d5c22cfc0c6/pre_processing/orac_preproc.F90#L483) so the override happens around where the driver file was read.

adamcpovey (changes requested):
There's a little more needed

pre_processing/rttov_driver.F90
garethethomas: You are, as usual, correct. I've updated accordingly - nevals is no more.

pre_processing/rttov_driver.F90
garethethomas: As with `rttov_driver.F90`, `nevals` is no more.

There doesn't seem to be much difference between `rttov_driver.F90` and `rttov_driver_gfs.F90` - perhaps a conversation about whether it is needed could be had....

pre_processing/set_ecmwf.F90
garethethomas: Fair enough. I've moved it to `orac_preproc.F90`, as suggested (i.e. your second option).

adamcpovey (approved):

PR #95, opened 2024-02-27
Significant restructuring of the post-processor code.

garethethomas:
The post-processor code is a bit of a mess. This update addresses some of the most obvious problems/confusing aspects of the code, while adding some useful new functionality:
(a) Code should be generally more readable, better commented and easier to understand now.
(b) Removed duplication of Bayesian selection code (and depreciated USE_NEW_BAYESIAN_SELECTION option).
(c) Only one input file worth of data is stored in memory at a time, regardless of which selection approach is used. 
(d) Added the ability to run both Bayesian (for aerosol) and Pavalonis (for cloud) selections (for combining both cloud and aerosol files into a single product). 
(e) Added the option to apply a simple weighting to the Bayesian selection method. The weights are passed via the BAYESIAN_WEIGHTS option in the driver file, and are treated as an a priori probability for each phase/type.


post_processing/orac_postproc.F90
```diff
@@ -329,6 +341,15 @@ subroutine orac_postproc(mytask, ntasks, lower_bound, upper_bound, &
        case('CORRECT_PARALLAX')
           if (parse_string(value, corr_plx) /= 0) &
                call handle_parse_error(label)
+       case('BAYESIAN_WEIGHTS')
+          ! Note that we expect the bayesian_weights array to have more elements
+          ! than given in the driver file!
+          stat = parse_string(value, bayesian_weights)
```
adamcpovey, line 347: Do we want to check that this is the same length as the number of files provided? It seems easy is accidentally add/subtract a single number from a 20 element list and unintentionally delete an aerosol type.

post_processing/orac_postproc.F90
```diff
@@ -492,54 +520,43 @@ subroutine orac_postproc(mytask, ntasks, lower_bound, upper_bound, &
...
-                     input_secondary(i))
-                call read_input_secondary_class(in_files_secondary(i), &
-                     input_secondary(i), loop_ind(i), chunk_starts(i_chunk), &
-                     verbose)
-             end if
-          end do
-       else ! Use traditional selection method
-          ! Load only the cost values from input files
-          do i = 1, n_in_files
-             call alloc_input_data_only_cost(loop_ind(i), input_primary(i), &
```
adamcpovey, line 516: I think the `_only_cost` routines are no longer called after this commit. In the name of tidiness, I'd be inclined to remove them, and get rid of the corresponding argument from `read_input_primary_class()`

post_processing/orac_postproc.F90
```diff
@@ -672,95 +648,88 @@ subroutine orac_postproc(mytask, ntasks, lower_bound, upper_bound, &
...
-                   input_primary(0)%phase(i,j) = IPhaseClU
+                ! Now apply cloud_cci specific changes
+                if (cloud_cci) then
+                   ! Overwrite cc_total with cldmask
+                   input_primary(0)%cc_total(i,j) = input_primary(0)%cldmask(i,j,1)
+                   input_primary(0)%cc_total_uncertainty(i,j) = &
+                        input_primary(0)%cldmask_uncertainty(i,j,1)
+                   ! Set non-cloud pixels to Pavalonis clear/unknown phase, if
+                   ! they don't already have an aerosol phase assigned.
+                   if (input_primary(0)%cldmask(i,j,1) .eq. 0 .and. &
```
adamcpovey, line 671: I don't remember why we do this. Oli implemented it.

post_processing/orac_postproc.F90
garethethomas: Is there any easy way of getting parse_string to tell you how many values it read? If not, this is not actually entirely trivial to do. If you did get the list of weights wrong, you wouldn't delete an aerosol type, since the weights default to 1, but you would have the wrong weights (obviously!)

post_processing/orac_postproc.F90
garethethomas: I guess it is because cc_total (the cloud fraction) is always either 1 or 0, as is determined by the cloud mask (?). Anyway, it might as well be left as it is for this pull request, so as to be consistent. Possibly something to think about changing separately though.

post_processing/orac_postproc.F90
garethethomas: Agreed.

post_processing/orac_postproc.F90
adamcpovey: Oh, wait, actually that's a bug (I didn't test this keyword).

`parse_string` takes an optional argument `n_in`, which tells it how long the array it reads should be. If that argument isn't given, the desired length is assumed to be length of the input array. In this case, that's `MaxInFiles` which won't be appropriate.

I think this needs to be,
```
stat = parse_string(value, bayesian_weights, n_in=n_in_files)
```
or
```
stat = parse_string(value, bayesian_weights(1:n_in_files))
```
whichever you think is easier to understand.

PR #96, opened 2024-03-18
Added the option to write the measurement errors to the secondary output file

garethethomas:
It can be quite useful to see what values for measurement error are actually being used in the retrieval, so this adds the option to request the diagonals of the Sy covariance matrix be output to the seondary.nc file.

### DIFF
--- a/src/get_spixel.F90
+++ b/src/get_spixel.F90
@@ -192,6 +192,9 @@
 !    surface reflectance.
 ! 2022/01/27, GT: Added MSI_Data as an argument to Get_X (needed if using
 !    SelmAUX to set a priori/first guess)
+! 2023/10/01, GT: Added some initial zeroing of channel counts, so that if the
+!    current pixel is skipped, it is apparent that SPixel data is not valid for
+!    the current pixel.
 !
 ! Bugs:
 ! None known.
@@ -228,6 +231,14 @@ subroutine Get_SPixel(Ctrl, SAD_Chan, SAD_LUT, MSI_Data, RTM, SPixel, status)
 
    SPixel%Type = MSI_Data%Type(SPixel%Loc%X0, SPixel%Loc%Y0)
 
+   ! Redefine some key SPixel%Ind values here, so that _if_ the current
+   ! pixel is skipped, it is apparent that the data contained within
+   ! SPixel is not valid for the current pixel
+   SPixel%Ind%Ny = 0
+   SPixel%Ind%NSolar = 0
+   SPixel%INd%NThermal = 0 
+   SPixel%Ind%NMixed = 0
+
    if (Ctrl%NTypes_to_process > 0) then
       if (.not. any(Ctrl%Types_to_process(1:Ctrl%NTypes_to_process) == &
                     SPixel%Type)) then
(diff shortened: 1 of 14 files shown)


src/get_spixel.F90
```diff
@@ -342,6 +354,16 @@ subroutine Get_SPixel(Ctrl, SAD_Chan, SAD_LUT, MSI_Data, RTM, SPixel, status)
    call Get_X(Ctrl, SPixel, MSI_Data, status)
 !  if (status /= 0) go to 99 ! Skip further data reading
 
+   ! It seems that it is possible that an invalid pixel, with no
+   ! defined measurement, can reach this point without causing a
+   ! SPixel status. Check for this condition here
+   if (SPixel%Ind%Ny == 0) then
```
adamcpovey, line 360: While I'm not surprised this is possible, someone should track down what the failure mode is and fix that too.

src/prepare_output_secondary.F90
```diff
@@ -330,6 +331,44 @@ subroutine prepare_output_secondary(Ctrl, i, j, MSI_Data, SPixel, Diag, &
...
+      !   allocated.
+      ! * If the Sy array has the correct dimension. If we have missing
+      !   channels and skipped pixels, it is also possible at Ny ends up
+      !   being larger than Sy, at this point in the code.
+      if (SPixel%Ind%Ny .gt. 1 .and. size(SPixel%Sy) .gt. 1 .and. &
+           SPixel%Ind%Ny*SPixel%Ind%Ny .le. size(SPixel%Sy)) then
+         do k=1,SPixel%Ind%Ny
+            !write(*,"(e9.2)", advance="no") sqrt(SPixel%Sy(k,k))
+            call prepare_short_packed_float( &
+                 sqrt(SPixel%Sy(k,k)), output_data%Sy(i,j,k), &
```
adamcpovey, line 355: This doesn't map channels between SPixel and Ctrl and so is outputting in the wrong place. I'll work out a fix momentarily.

adamcpovey (changes requested):
I'll make some whitespace tidying and work out the channel indexing

adamcpovey:
After 20 minutes I worked out how to push changes to your branch and update this thing! @garethethomas please check that those still address the problems you were having (I went looking for the causes of the missed status flag rather than circumvent the whole system).

Further, I am beginning to question a bunch of stuff in here:

- Why are we allocating SPixel arrays each time when they just get copied into local arrays in `invert_marquadt()`? We probably save a bunch of processing time by not bothering to reallocate.
- Given the output arrays are initialised to fill values, why do we call `prepare_output_XXX()` for skipped retrievals? Surely doing nothing outputs what we want (namely nothing)?

garethethomas:
I agree with your inclusion of `SPixel%spixel_y_to_ctrl_y_index` - from what I've seen it isn't generally necessary (it's a one-to-one match), but I guess this might not _always_ be the case!
Regarding your other change - it _should_ be ok, so long as `SPixel%Ny = 0` means that the do loop at line 338 of prepare_output_secondary.F90  doesn't get executed. Is that how Fortran works?

I was also surprised at the fact that all of the SPixel arrays are reallocated for every pixel - like you say, it seems rather inefficient. "Fixing" this would be a big job though!

adamcpovey:
`SPixel%spixel_y_to_ctrl_y_index` is mostly important for the cloud retrieval, as one of the effective radius channels is _always_ missing when multiple are present.

To my knowledge, Fortran checks the validity of the for condition at the start of the loop so `do i=1,0` should immediately terminate.

adamcpovey:
I never managed to work out how to get sensible errors out of flex. You can get it to at least throw a line number and slightly more information (almost always "syntax", but others can happen) by uncommenting the last two lines of `config/arch.inc`

PR #100, opened 2024-04-19
New uncertainty characterisation updates

garethethomas:
Added a Ctrl variable for switching between the new measurement uncertainty characterisation (if using the NetCDF LUTs) and the old approach. 
Also added deallocation of Sy output arrays, and removed unnecessary scale_factor and offset attributes from integer NetCDF variables.

### DIFF
--- a/common/def_output_primary.F90
+++ b/common/def_output_primary.F90
@@ -1582,8 +1582,6 @@ end if
            long_name     = 'number of retrieval iterations', &
            standard_name = '', &
            fill_value    = byte_fill_value, &
-           scale_factor  = output_data%niter_scale, &
-           add_offset    = output_data%niter_offset, &
            valid_min     = output_data%niter_vmin, &
            valid_max     = output_data%niter_vmax, &
            units         = '1', &
@@ -1710,8 +1708,6 @@ end if
            long_name     = 'land/sea flag', &
            standard_name = 'land_binary_mask', &
            fill_value    = byte_fill_value, &
-           scale_factor  = output_data%lsflag_scale, &
-           add_offset    = output_data%lsflag_offset, &
            valid_min     = output_data%lsflag_vmin, &
            valid_max     = output_data%lsflag_vmax, &
            units         = '1', &
@@ -1732,8 +1728,6 @@ end if
            long_name     = 'land use flag', &
            standard_name = 'land_use_mask', &
            fill_value    = byte_fill_value, &
-           scale_factor  = output_data%lusflag_scale, &
-           add_offset    = output_data%lusflag_offset, &
            valid_min     = output_data%lusflag_vmin, &
            valid_max     = output_data%lusflag_vmax, &
            units         = '1', &
@@ -1801,8 +1795,6 @@ end if
            long_name     = 'illumination flag', &
            standard_name = '', &
            fill_value    = byte_fill_value, &
-           scale_factor  = output_data%illum_scale, &
-           add_offset    = output_data%illum_offset, &
            valid_min     = output_data%illum_vmin, &
            valid_max     = output_data%illum_vmax, &
            units         = '1', &
@@ -1834,8 +1826,6 @@ end if
            long_name     = 'Pavolonis cloud type', &
            standard_name = '', &
            fill_value    = byte_fill_value, &
-           scale_factor  = output_data%cldtype_scale, &
-           add_offset    = output_data%cldtype_offset, &
            valid_min     = output_data%cldtype_vmin, &
            valid_max     = output_data%cldtype_vmax, &
            units         = '1', &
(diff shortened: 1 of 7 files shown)


common/def_output_secondary.F90
```diff
@@ -750,9 +750,17 @@ subroutine def_output_secondary(ncid, dim3d_var, output_data, indexing, &
          input_dummy2 = 'measurement uncertainty in channel no '// &
               trim(adjustl(input_num))
          if (btest(indexing%Ch_Is(i), ThermalBit)) then
-            input_dummy3 = 'kelvin'
+            input_dummy='measurement_uncertainty_in_channel_no_'// &
+                 trim(adjustl(input_num))
+            input_dummy2='measurement uncertainty in channel no '// &
+                 trim(adjustl(input_num))
+            input_dummy3='kelvin'
```
adamcpovey, line 757: Don't these just repeat the block immediately before the `if`?

common/def_output_primary.F90
```diff
@@ -1582,8 +1582,6 @@ subroutine def_output_primary(ncid, dim3d_var, output_data, indexing, &
            long_name     = 'number of retrieval iterations', &
            standard_name = '', &
            fill_value    = byte_fill_value, &
-           scale_factor  = output_data%niter_scale, &
-           add_offset    = output_data%niter_offset, &
```
adamcpovey, line 1586: Good idea. Should also delete them from the output_data structure.

src/read_sad_lut.F90
```diff
@@ -1225,8 +1228,12 @@ subroutine Read_NCDF_SAD_LUT(Ctrl, LUTFilename, SAD_Chan, SAD_LUT)
             ! Uncertainty = a**2 L**2 + b**2 L + c**2
             call ncdf_read_array(fid, "ru"//char(j+96), chan_tmp_real)
             do i = 1, Ctrl%Ind%NSolar
+               ! ru values stored in LUT already squared?!
                SAD_Chan(Ctrl%Ind%YSolar(i))%Solar%ru2(j) = &
-                    chan_tmp_real(solar_indices(i)) ** 2
+                    chan_tmp_real(solar_indices(i)) !** 2
+               ! Ensure SNR is explicitly set to 0 (used to determine if ru2
+               ! should be used in get_measurments)
+               SAD_Chan(Ctrl%Ind%YSolar(i))%Solar%SNR = 0.0
```
adamcpovey, line 1236: Isn't this done on [line 1198](https://github.com/ORAC-CC/orac/blob/master/src/read_sad_lut.F90#L1198)?

src/get_measurements.F90
```diff
@@ -199,16 +203,17 @@ subroutine Get_Measurements(Ctrl, SAD_Chan, SPixel, MSI_Data, status)
...
                SPixel%Sy(i,i) = SAD_Chan(ii)%Solar%NEDR
             else
                ! New LUTs approach (varying uncertainty)
                ! Convert reflectance to radiance
                j = SPixel%ViewIdx(i)
-               Lx = (SPixel%Ym(i) * cos(SPixel%Geom%SolZen(j) * d2r) * &
-                     SAD_Chan(ii)%Solar%F0) / Pi
+               !Lx = (SPixel%Ym(i) * cos(SPixel%Geom%SolZen(j) * d2r) * &
+               !      SAD_Chan(ii)%Solar%F0) / Pi
+               Lx = (100.0 * SPixel%Ym(i) * SAD_Chan(ii)%Solar%F0) / Pi
```
adamcpovey, line 216: Add a comment explaining why we're removing the \cos\theta. We won't remember next time we go to war with The Angles.

src/read_sad_lut.F90
```diff
@@ -1225,8 +1228,12 @@ subroutine Read_NCDF_SAD_LUT(Ctrl, LUTFilename, SAD_Chan, SAD_LUT)
             ! Uncertainty = a**2 L**2 + b**2 L + c**2
             call ncdf_read_array(fid, "ru"//char(j+96), chan_tmp_real)
             do i = 1, Ctrl%Ind%NSolar
+               ! ru values stored in LUT already squared?!
```
adamcpovey, line 1231: The units in the files agree with you here. I shall check what Don meant.

adamcpovey (changes requested):
This compiles and runs without regression on old tables. (I'll check new tables next.) A few notes for our future selves would be helpful.

src/read_sad_lut.F90
adamcpovey: Don says the values are not squared and so they need to be here.

adamcpovey:
Minor prod of @garethethomas to add some comments to the code so we can accept this PR.

PR #102, opened 2024-10-15
Adding functionality to use ECMWF forecast or era5 grid instead of pre-processing grid + additional dustflag from Gareth

dhegedus99:
The main aim of this PR is to eliminate the need to use the emos library by interpolating onto a pre-defined pre-processing grid. The new edits automatically use the ECMWF forecast/era5(t) grids for the pre-processing. In order to do this, I changed the indexing that is used for the preproc_structures variables, as the x- and y-dimensions were using the start and end pixel numbers to define the arrays, while the k-dimension used 1 to kdim. I've changed it so all dimensions have array indices between 1 and the length of the dimension (xdim, ydim, kdim) for simplicity. I haven't enabled it by default, but if you set USE_ECMWF_PREPROC_GRID=True as one of the arguments for the pre-processor, then it will not use emos to interpolate.

Another minor change in this PR is Gareth's dust flag, which is to help reduce the misclassification of thick dust as cloud in the pre-processor.


adamcpovey:
@simonrp84 Yeah, I'm looking at it now. It compiles, just checking nothing breaks and aiming to add the flags for EMOS to be truly optional.

common/def_output_primary.F90
```diff
@@ -1823,7 +1825,10 @@ subroutine def_output_primary(ncid, dim3d_var, output_data, indexing, &
                'opaque_ice ' // &
                'cirrus ' // &
                'overlap ' // &
-               'prob_opaque_ice'
+               'prob_opaque_ice ' // &
+               'N/A ' // &
```
adamcpovey, line 1829: Why isn't this `prob_clear`?

common/morphology.F90
```diff
@@ -0,0 +1,217 @@
...
+! None (see above)
+!
+! History:
+! 2024/03/08, GT: First version
+!
+! Bugs:
+! None known
+!-------------------------------------------------------------------------------
+
+module morphology_m
```
adamcpovey, line 40: I note that the OpenCV library implements these functions. Someone should, one day, check how much more efficient they are than this.

pre_processing/correct_for_ice_snow.F90
```diff
@@ -649,15 +652,12 @@ subroutine correct_for_ice_snow_nwp(nwp_path, imager_geolocation, &
...
-         lon_i = floor((imager_geolocation%longitude(i,j) + &
-              preproc_dims%lon_offset)*preproc_dims%dellon, kind=lint) + 1
-         lat_j = floor((imager_geolocation%latitude(i,j) + &
-              preproc_dims%lat_offset)*preproc_dims%dellat, kind=lint) + 1
-
-         if (lon_i .lt. preproc_dims%min_lon) lon_i = preproc_dims%min_lon
-         if (lat_j .lt. preproc_dims%min_lat) lat_j = preproc_dims%min_lat
-         if (lon_i .gt. preproc_dims%max_lon) lon_i = preproc_dims%max_lon
-         if (lat_j .gt. preproc_dims%max_lat) lat_j = preproc_dims%max_lat
+         lat_j = minloc(abs(preproc_geoloc%latitude - imager_geolocation%latitude(i,j)),1)
```
adamcpovey, line 655: This seems inefficient. Don't we know that the preproc_geoloc grid is going to be regular? The interpolation of the RTTOV outputs assumes a rectilinear grid so we should be able to assume that here too and use basically the same calculation as before?

pre_processing/get_surface_emissivity.F90
```diff
@@ -382,41 +383,38 @@ subroutine get_camel_emissivity(cyear, cmonth, camel_emis_path, imager_flags, &
...
+   allocate(counter(1:preproc_dims%xdim, 1:preproc_dims%ydim))
+   allocate(summat(1:preproc_dims%xdim, 1:preproc_dims%ydim, n_chans))
 
    counter = 0
    summat  = 0.
    do j = 1, emis%nlat
-      lat = floor((emis%lat0+(j-1)*emis%lat_del+preproc_dims%lat_offset)* &
-            preproc_dims%dellat)+1
-      if (lat.ge.preproc_dims%min_lat .and. lat.le.preproc_dims%max_lat) then
+      if ((emis%lat0+(j-1)*emis%lat_del).ge.minval(preproc_geoloc%latitude) .and. &
```
adamcpovey, line 399: Isn't `minval(preproc_geoloc%latitude)` equal to `preproc_dims%min_lat`?

adamcpovey (approved):
This compiles and runs for me, with no regressions using the previous processing style. Using the ERA5 grid introduces bigger changes in `stemp` then I expected, which warrants some investigation, but I'm happy to merge this once Dora confirms I haven't broken anything with my tidying.

common/def_output_primary.F90
dhegedus99: I think I took this from Gareth's original implementation of the dustflag, and didn't realise it corresponds to prob_clear - I can change it to "prob_clear"

pre_processing/get_surface_emissivity.F90
dhegedus99: It isn't, the name of it is misleading - I think it is meant to refer to the index of the minimum latitude. When using the old preprocessing grid, the the array of latitudes and longitudes are calculated on this line: https://github.com/ORAC-CC/orac/blob/4d55dbb924017fcccbed3c03dd130d0ba9d96f2c/pre_processing/define_preprop_grid.F90#L59
So for a minimum latitude of 41.4degrees, the preproc_dims%min_lat is equal to 183 (of integer type). I think this is the index of the minimum latitude on the preprocessing grid, instead of being the actual minimum value. 
When using the ECMWF native grid, `minval(preproc_geoloc%latitude)` is almost equal to `preproc_dims%min_lat`, but the latter is an integer value of the minimum latitude. I have instead added an additional preproc_dims%min_lat_ind variable that corresponds to the index of the minimum latitude value in the ECMWF grid.
I have basically tried to remove the use of `preproc_dims%min_lat` in other parts of the pre-processor because of the diverging/misleading meaning of this variable.

pre_processing/correct_for_ice_snow.F90
dhegedus99: This code is looping through the imager pixels, whereas rttov was looping through the preprocessing grid points, so I couldn't quite simplify it as much as in the rrtov code. Have you got any suggestions?

common/morphology.F90
simon-sat: My version of this code in python implements the OpenCV versions and they're pretty fast. Can't do direct comparison as I never had the time but I suspect the runtime is, at worst, equal. From memory, the OpenCV ones are C# code wrapped in python.

common/def_output_primary.F90
dhegedus99: In the main ORAC repo it is also written as prob_opaque_ice. Should I change or resolve this conversation and merge?

